Issue #29, opened 2020-11-13
Gather the Use Case and Requirements for RDF*


afs:
This issue is to collect together the use cases for RDF*.

Please add use cases you are aware of as comments below.

The document started in EasierRDF has been [copied into this repo](/w3c/rdf-star/tree/main/UCR) and is visible as HTML at:

https://w3c.github.io/rdf-star/UCR/

Recording a UC&R is not a promise by the CG to address that use case.

A useful format is:

>     As an <actor>       … WHO
>     I want a <feature>  … WHAT
>     So that <benefit>   … WHY

gatemezing:
As a data engineer,
I want to add meta-properties over properties, from an existing SQL-DB to a Graph-DB; that is for example
_for a node of type PERSON, with a property :birthDate='1985-01-20', I want to be able to say 
the property was created on '2020-11-10', by a user 'editor1' with the role 'writer'._


So that I do not need to transform the property into a relation+ yet another node.

HolgerKnublauch:
As an ontologist/schema developer using SHACL
I want to specify sh:message and sh:severity for individual constraints, e.g. attach to individual sh:minCount 1 triples
because otherwise I would have to create an individual shape (subject) just for this information.

(This use case is already implemented in our TopBraid EDG product)

lisp:
As a data service, from an architectural perspective, interpretation should be distinct from and cleanly layered over representation - both abstract and concrete, in order to facilitate implementation and assure the ability to support future use cases as knowledge evolves over time.

From this perspective, the standard treatment of blank node labels is an anti-requirement, the lessons of which should dissuade from unnecessarily restrictive relations between interpretation and representation.

akuckartz:
> the standard treatment of blank node labels is an anti-requirement, the lessons of which should dissuade from unnecessarily restrictive relations between interpretation and representation.

Could you please explain the use case in a way I can immediately understand? Thanks ;-)

lisp:
the situation with blank node labels is a case where the interpretation was fixed, to treat them as referentially opaque. that was a mistake. it would have been better to separate the interpretation from the representation and to allow alternatives. 

- https://lists.w3.org/Archives/Public/semantic-web/2009Nov/0040.html
- https://www.slideshare.net/PatHayes/blogic-iswc-2009-invited-talk
- http://videolectures.net/iswc09_hayes_blogic/ (which, just now, did not work)

akuckartz:
[...]

PR #57, opened 2020-12-08
update links to GH issues

pchampin:
issue #29 is not the only one related to UCS anymore

### DIFF
--- a/UCR/rdf-star-ucr.html
+++ b/UCR/rdf-star-ucr.html
@@ -43,26 +43,14 @@
         <a href="https://www.w3.org/community/rdf-dev/">W3C RDF Community Group</a>
         on the <a href="https://lists.w3.org/Archives/Public/public-rdf-star/"
                   >RDF* mailing list</a>
-        and <a href="https://github.com/w3c/rdf-star/issues/29">issue 29</a>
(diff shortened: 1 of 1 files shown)


UCR/rdf-star-ucr.html
```diff
@@ -43,26 +43,14 @@ <h2>Introduction</h2>
...
                   >RDF* mailing list</a>
-        and <a href="https://github.com/w3c/rdf-star/issues/29">issue 29</a>
+        and <a href="https://github.com/w3c/rdf-star/issues?q=is%3Aissue+is%3Aopen+label%3Ause-case">issues</a>
         of the  community's GitHub repository.
     </section>
 
     <section>
       <h2>Contributing Use Cases and Requirements</h2>
       <p>
-        Add a comment to
```
afs, line 53: My preference is leave there text about WHO-WHAT-WHY and also have the template link. i.e. Explanation and indication what is about to happen and the active link.

afs (approved):
Comment made but not a blocker.

UCR/rdf-star-ucr.html
pchampin: I just put it back

Issue #20, opened 2020-10-19
Rename "RDF*" - avoiding regular expression wildcard in name


akuckartz:
Copied from https://github.com/w3c/EasierRDF/issues/76

> The name "RDF*" has the negative property that it is not just a string, but also a regular expression. Some search engines make it difficult or impossible to search for "RDF*" without interpreting the name as a regular expression.
> 
> Some suggestions:
> 
> * RDF star
> * RDFx

@hartig commented in https://github.com/w3c/EasierRDF/issues/76#issuecomment-708232084
> I don't think it is a good idea to rename RDF* at this point. However, I understand the issue related to search engines. A possible way to address this issue is to include keywords such as "RDF star", "RDFstar", "SPARQL star", etc, in the metadata sections of the documents about the approach; this way, search engine can pick up these keywords and, then, searches for these keywords will end up showing the right hits.

I think it is not to late to rename the specification by replacing "RDF*" with "RDF star". Also:
* Quite a few articles about `RDF*` already write `RDF star` instead
* This repository is named `rdf-star`
* html metadata keywords are almost never provided on the web today and it is questionable that search engines use them
* html metadata keywords can not be provided for PDF files etc.

jeenbroekstra:
Freewheeling here: how about "RDF edge properties"?

HolgerKnublauch:
It probably depends on how close the outcome of this work resembles the original RDF* proposals. RDF* is now an established term but with this comes a long history and that may eventually be misleading.

I do like the idea of naming it after what it really does, as Jeen's Edge Properties (which is a bit long though).

gatemezing:
I still go for RDF* for the name. However, we can find ways to solve issues with the search engines by adding as much as possible alternative keywords/structured data so as to help them "understand" better.

VladimirAlexiev:
Rename it to  "RDF star". Now is the time.
- Just because RDF* is used in papers does not mean we should perpetuate that unfortunate name . Search for `RDF*` in Google scholar and Semantic scholar and you'll see what I mean
- we can't use `*` in Mime types so that's another reason

"Edge properties" does not work. What will we say,  "sparql with edge properties" and "sparql results XML with edge properties"?  Come on

HolgerKnublauch:
Historical question: what was the reason/meaning behind the * anyway?

Wouldn't RDF plus be a more meaningful alternative?

hartig:
[...]

PR #61, opened 2020-12-10
Update index.html

TallTed:
added some meeting details, made "RDF-Star" the primary name

### DIFF
--- a/index.html
+++ b/index.html
@@ -50,8 +50,8 @@
         </div> 	
 
         <p>
-          <a href="https://w3c.github.io/rdf-star/">RDF* Home</a>, part of the 
-          <a href="https://www.w3.org/community/rdf-dev/">RDF dev community group.</a>
+          <a href="https://w3c.github.io/rdf-star/">Home of RDF-Star (née RDF*)</a>, a part of the 
+          <a href="https://www.w3.org/community/rdf-dev/">RDF-DEV community group.</a>
         </p>
 
         <p>Community:</p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -50,8 +50,8 @@ <h1>RDF*</h1>
         </div> 	
 
         <p>
-          <a href="https://w3c.github.io/rdf-star/">RDF* Home</a>, part of the 
-          <a href="https://www.w3.org/community/rdf-dev/">RDF dev community group.</a>
+          <a href="https://w3c.github.io/rdf-star/">Home of RDF-Star (née RDF*)</a>, a part of the
```
pchampin, line 53: I generally agree that a latinized version of "RDF*" is required in all related documents, including this page. However, I am not sure that "RDF Star" becoming the primary name would make consensus. Something like

>  Home of RDF* (pronounced "RDF star")

would be less contreversial at the moment, IMHO

hartig: That's an excellent compromise.

TallTed: fine...

TallTed, line 53:
```suggestion
          <a href="https://w3c.github.io/rdf-star/"> Home of RDF* (pronounced "RDF star")</a>, a part of the 
```

TallTed:
Accidentally deleted fork; recreating in new PR

index.html
akuckartz: > That's an excellent compromise.

I am not convinced that this is a "compromise" (let alone an "excellent" one).

In any case it does not resolve #20

TallTed: +1 @akuckartz.  That said, as this was just meant to improve the homepage, I'm OK with it not resolving #20.

PR #68, opened 2020-12-12
Turtle evaluation tests

afs:
Includes

- Turtle evaluation tests, include annotation tests
- N-Triples* syntax tests
- Every NT* syntax test is copied, and in the NT* manifest as a Turtle syntax test

### DIFF
--- a/tests/turtle/eval/manifest.ttl
+++ b/tests/turtle/eval/manifest.ttl
@@ -1,0 +1,79 @@
+## Distributed under both the "W3C Test Suite License" [1]
+## and the "W3C 3-clause BSD License".
+## [1] https://www.w3.org/Consortium/Legal/2008/04-testsuite-license
+## [2] https://www.w3.org/Consortium/Legal/2008/03-bsd-license
+
+PREFIX rdf:    <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+PREFIX rdfs:   <http://www.w3.org/2000/01/rdf-schema#>
+PREFIX mf:     <http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#>
+PREFIX rdft:   <http://www.w3.org/ns/rdftest#>
+PREFIX :       <#>
+
+<>  rdf:type mf:Manifest ;
+    rdfs:label "Turtle* Evaluation Tests" ;
+    mf:entries
+    (
+        :turtle-star-1
+        :turtle-star-2
+        :turtle-star-bnode-1
+        :turtle-star-bnode-2
+        :turtle-star-annotation-1
+        :turtle-star-annotation-2
+        :turtle-star-annotation-3
+        :turtle-star-annotation-4
+        :turtle-star-annotation-5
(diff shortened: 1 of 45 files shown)


tests/turtle/eval/manifest.ttl
```diff
@@ -0,0 +1,79 @@
...
+## [2] https://www.w3.org/Consortium/Legal/2008/03-bsd-license
+
+PREFIX rdf:    <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+PREFIX rdfs:   <http://www.w3.org/2000/01/rdf-schema#>
+PREFIX mf:     <http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#>
+PREFIX rdft:   <http://www.w3.org/ns/rdftest#>
+PREFIX :       <#>
+
+<>  rdf:type mf:Manifest ;
+    rdfs:label "Turtle* EvaluationSyntax" ;
```
gkellogg, line 13: ```suggestion`
    rdfs:label "Turtle* Evaluation Tests" ;
```

gkellogg (changes requested):
Some of the NTriples test files referenced from the manifest differ from the actual test files:

* `ntriples-star-syntax-bad-*.nt` vs `ntriples-star-bad-syntax-*.nt`
* `nt-ttl-star-syntax-bad-*.nt` vs `nt-ttl-star-bad-syntax-*.nt`

Also, please update /index.html with references to these test manifests; HTML version, if it's generated.

Otherwise, I pass all the tests.

tests/turtle/eval/manifest.ttl
afs: Changes made.  Thanks for catching the errors.

afs:
> Also, please update /index.html with references to these test manifests; HTML version, if it's generated.

I have tried to do that. Added to this PR.

@pchampin - I had to change the `make-html-manifests.py` to remove the validation of the id name because it was checking for `#turtle-star-`.

pchampin:
> @pchampin - I had to change the make-html-manifests.py to remove the validation of the id name because it was checking for #turtle-star-.

No problem... I initially added that to detect copy-paste errors (copying a test and forgetting to change its name or its id) but I should probably run this check in a separate script...

gkellogg (approved):
Thanks, but still no link to the Turtle tests from /index.html. Otherwise, LGTM.

afs:
HTML rebuilt now; links on home page (and fixed the insecure links warning).

pchampin:
Thanks a lot :)
One thing I believe is missing: a test case for Turtle* and N-Triples* where there is no space between `<<` and the embedded subject. I think I encountered some implementations that didn't support it, but my understanding of the grammar is that it should (even though that's not very readable...). Do you agree?

afs:
It is possible.

`<<<uri> :q 123>>` and `<<<<:s :p :o>> :q :z>>` and similarly the trailing case.

I'll add some tests for this.

PR #70, opened 2020-12-14
More syntax tests

afs:
Includes a bad syntax test for empty annotation.

### DIFF
--- a/tests/nt/syntax/manifest.html
+++ b/tests/nt/syntax/manifest.html
@@ -54,9 +54,11 @@
 <li class="proposed"><a href="#ntriples-star-1">N-Triples* - subject triple term</a>
 <li class="proposed"><a href="#ntriples-star-2">N-Triples* - object triple term</a>
 <li class="proposed"><a href="#ntriples-star-3">N-Triples* - subject and object triple terms</a>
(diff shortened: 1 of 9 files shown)


afs:
@gkellogg

gkellogg:
@afs, only committers to the repo can be added as reviewers, which I apparently, am not. However, I can submit a review.

afs:
Just to check the structure - the details of the tests are - obviously - for the whole community.

The interesting bit is the bad syntax test for empty annotation.

For a while, it thought "symmetry" (= tools generation) but realised that syntax that generated nothing was not useful.

gkellogg (approved):

gkellogg:
It's consistent with `predicateObjectList` everyplace other than after a `blankNodePropertyList`, otherwise, we'd need `"{|" predicateObjectList? "|}"` which does, indeed, seem pointless.

josd:
+1 and EYE is currently throwing an empty_triple_annotation exception
The other test results are at https://github.com/josd/eye/blob/master/reasoning/rdf-star/run-outcome.n3

BTW @afs thanks for riot without which we would not be able to run https://github.com/josd/eye/blob/master/reasoning/rdf-star/run-test

PR #72, opened 2020-12-15
Add telecon time to home page.

afs:


TallTed:
Seems to partially duplicate #61

afs:
Yes - this is just about the telecons and about not rewriting other text.

TallTed:
I don't understand why you would create a new PR for that, instead of suggesting changes to the existing PR which changes the same snippet.

afs:
Because #61 is trying to change the name of the entire endeavour and that is obstructing making this small, unrelated change.

TallTed:
#61 is gone, due to an accidental fork deletion.

#74 revises the text, to not rename but still incorporate the "rdf star" lexical, so that clever web searchers who figure out that they should replace `*` with `star` have a chance of finding it.

#74 also includes the meeting details that were ignored in #61 because the naming issue was so blinding.

pchampin:
superceded by #74

Issue #64, opened 2020-12-11
are embedded triples unique?


pfps:
The uniqueness of
embedded triples does not appear to me to be determined - the initial RDF* and
the current documents say yes in their formal sections but central examples
have fatal flaws if embedded triples are unique.

These foundational issues should be cleared up early.

TallTed:
The meaning of your title is not obvious to people not versed in some discipline, in which I am also not versed so cannot even name.

I have come to understand your title means "Do embedded triples always have the same meaning?" i.e., "Does each embedded triple lexical have a singular meaning, across all instances of that triple lexical, regardless of graph, embeddedness, etc?"

I think the answer is "Yes, unless blank nodes are involved; then, the answer is indeterminate."

pchampin:
I didn't want to pollute this discussion with a looong comment, but FTR
> central examples have fatal flaws if embedded triples are unique

I explained in details on the mailing list why I think said examples are imperfect, but not fatally flawed
https://lists.w3.org/Archives/Public/public-rdf-star/2020Dec/0076.html

pchampin:
This was discussed during today's call: https://w3c.github.io/rdf-star/Minutes/2020-12-18.html#item04

TallTed:
Seems that distinction should be made between blank node `[]` and blank nodes like `_:a`, as the former are treated as different nodes at each occurrence (even if some or all are externally known to have `owl:sameAs` relations), while the latter are treated as the same node in all occurrences within the same (named) graph and/or within the same Turtle/N3 document.  We cannot speak simply of "blank nodes".

TallTed:
Also, in [today's minutes](https://w3c.github.io/rdf-star/Minutes/2020-12-18.html#item04) --
> <pfps> Well, there is nowhere the triple with subject :a, as triples can't have CURIES as subjects.

What forbids CURIES as subjects?

So far as I'm aware, CURIES (or more likely to be accurate in this context, prefixed names) are simply shorthand for URI/URL/IRIs, and as such are valid RDF (and RDF*) subjects.

What am I missing?

pfps:
Neither [] nor _:a are blank nodes.   The latter is a blank node label and the former is ANON, both part of the Turtle syntax.  Processing Turtle into RDF graphs turns both of them into blank nodes, of course, and in different ways.

pfps:
From RDF 1.1 Concepts and Abstract Syntax:

An RDF triple consists of three components:

    the subject, which is an IRI or a blank node
[...]

PR #75, opened 2020-12-17
Add material in the spec about the triple-occurrence distinction

pchampin:
Aims to addresses issue #64 by
* fixing misleading examples (`prov:wasDerivedFrom`)
* adding an informative section on the type-token distinction, and providing guidance on how to model occurrences with RDF*
* adding an apologetic appendix on the controversial seminal example (with @hartig 's permission :wink:)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/pull/75.html" title="Last updated on Feb 4, 2021, 5:24 PM UTC (8e3590f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/75/ce57841...8e3590f.html" title="Last updated on Feb 4, 2021, 5:24 PM UTC (8e3590f)">Diff</a>


pfps:
I am against using type-token for this distinction.   Consider URIs.  Are ex:a and ex:a two different tokens for the same type?  No!   <:a :b :c> and <:a :b :c> are not either nor are "1"^^xsd:int and "1"^^xsd:int. The RDF* documents should use wording that makes sense if used for IRIs or RDF literals.

rdf-star-cg-spec.html
```diff
@@ -173,6 +173,30 @@ <h2>Concepts and Abstract Syntax</h2>
     </ul>
 
     <p>Again, this definition is an extension of the notion of <dfn data-cite="RDF11-CONCEPTS#dfn-rdf-dataset">RDF dataset</dfn>, hence it follows that any <a>RDF dataset</a> is also an <a>RDF* dataset</a>.</p>
+
+    <section class="informative" id="type-token">
+      <h2>RDF* triples and the type-token distinction</h2>
+
+      <p>According to the definitions above, an <a>RDF* triple</a> is an abstract entity whose identity is entirely defined by its subject, predicate an object. In other words, given three <a>RDF* terms</a> |s|, |p| and |o|, the embedded triple (|s|, |p|, |o|) is assumed to represent the <em>same thing</em> everywhere it occurs, including across different graphs (just like the same IRI |p| is assumed to represent the same thing everywhere it occurs).</p>
```
TallTed, line 180:
```suggestion
      <p>According to the definitions above, an <a>RDF* triple</a> is an abstract entity whose identity is entirely defined by its subject, predicate and object. In other words, given three <a>RDF* terms</a> |s|, |p|, and |o|, the embedded triple (|s|, |p|, |o|) is assumed to represent the <em>same thing</em> everywhere it occurs, including across different graphs (just like the same IRI |p| is assumed to represent the same thing everywhere it occurs).</p>
```

rdf-star-cg-spec.html
```diff
@@ -173,6 +173,30 @@ <h2>Concepts and Abstract Syntax</h2>
...
+
+      <p>In some situations, however, it might be necessary to distinguish the <em>occurrences</em> of a triple (in different files, for example). Consider the following sentence: "The triple `:s :p :o` was added in file1.ttl by Alice, and the same triple was added in file2.ttl by Bob." Note that words "same triple" in this sentence may be confusing, because although the triple (as an abstract entity) is the same, its respective occurrences in the two files are different things, each within a different file and with a different author. This is known as the <a href="https://en.wikipedia.org/wiki/Type%E2%80%93token_distinction">type-token distinction</a>. As the embedded triple represents a unique abstract entity, adequately conveying the meaning of the sentence above requires additional nodes to represent each occurrence, as illustrated below.</p>
```
TallTed, line 182:
```suggestion
      <p>In some situations, however, it might be necessary to distinguish the <em>occurrences</em> of a triple (in different Turtle or Turtle* files, for example). Consider the following sentence: "The triple `:s :p :o` was added in file1.ttl by Alice, and the same triple was added in file2.ttl by Bob." Note that words "same triple" in this sentence may be confusing, because although the triple (as an abstract entity) is the same, its respective occurrences in the two files are different things, each within a different file and with a different author. This is known as the <a href="https://en.wikipedia.org/wiki/Type%E2%80%93token_distinction">type-token distinction</a>. As the embedded triple represents a unique abstract entity, adequately conveying the meaning of the sentence above requires additional nodes to represent each occurrence, as illustrated below.</p>
```

rdf-star-cg-spec.html
```diff
@@ -989,6 +1013,29 @@ <h2>SA-mode and PG-mode</h2>
...
+      >
+        <!--
+        # the controversial seminal example
+        :bob foaf:name "Bob".
+        <<:bob foaf:age 23>> dct:creator <http://example.com/crawlers#c1> ;
+                             dct:source <http://example.net/listing.html> .
+        -->
+      </pre>
+
+      <p>This example was further debated on the <a href="https://lists.w3.org/Archives/Public/public-rdf-star/">RDF* mailing list</a>, as it may have set wrong expectations about what RDF* embedded triples represent. More precisely, from this example, one may wrongly assume that `<<:bob foaf:age 23>>` represent the <em>occurrence</em> of the given triple at the address `http://example.net/listing.html`. This impression may be reinforced by the use of `dct:creator`: arguably, a triple (as a unique abstract entity) is not "created" by anyone, while an occurrence thereof can be said to be created of authored.</p>
```
TallTed, line 1033:
```suggestion
      <p>This example was further debated on the <a href="https://lists.w3.org/Archives/Public/public-rdf-star/">RDF* mailing list</a>, as it may have set wrong expectations about what RDF* embedded triples represent. More precisely, from this example, one may wrongly assume that `<<:bob foaf:age 23>>` represent the <em>occurrence</em> of the given triple at the address `http://example.net/listing.html`. This impression may be reinforced by the use of `dct:creator`: arguably, a triple (as a unique abstract entity) is not "created" by anyone, while an occurrence thereof can be said to be created or authored.</p>
```

rdf-star-cg-spec.html
```diff
@@ -989,6 +1013,29 @@ <h2>SA-mode and PG-mode</h2>
...
+
+      <p>In retrospect, it would probably have been better to only use the `dct:source` property on the triple, and maybe to give it two values (as it is possible for the same unique triple to be mentioned in several sources). Another solution would have been to adopt a more complex modeling as in <a href="#type-token-example"></a>. However, the purpose of that example was not to discuss the subtleties of the type-token distinction (see <a href="#type-token"></a>), but on the contrary to provide a simple minimal example of how RDF* could be used.</p>
```
TallTed, line 1035:
```suggestion
      <p>In retrospect, it would probably have been better to only use the `dct:source` property on the triple, and maybe to give it two values (as it is possible for the same unique triple to be mentioned in several sources). Another solution would have been to adopt a more complex modeling, as in <a href="#type-token-example"></a>. However, the purpose of that example was not to discuss the subtleties of the type-token distinction (see <a href="#type-token"></a>), but on the contrary to provide a simple minimal example of how RDF* could be used.</p>
```

rdf-star-cg-spec.html
pchampin: ah, oxford comma :)

pchampin:
> Consider URIs. Are ex:a and ex:a two different tokens for the same type?

Yes, if I read correctly https://plato.stanford.edu/entries/types-tokens/#WhaDis .

>     Rose is a rose is a rose is a rose.
> In one sense of ‘word’ we may count three different words; in another sense we may count ten different words. C. S. Peirce (1931-58, sec. 4.537) called words in the first sense “types” and words in the second sense “tokens”.

How many IRIs do you count in this?

    rdfs:Class rdf:type rdfs:Class;
               rdfs:subClassOf rdfs:Class.

I count 3 IRI types, and 5 IRI tokens.

I don't consider the terms of RDF (IRIs, literals...) and RDF* (+triples) to be different, in that respect, from the terms of the English language. 

Of course, we are *not* talking here about what the terms *denote* (as in the example in the link above: in "an 8,000 year old bean", does the word "bean" denote a bean type or a bean token?). We are talking about terms types and term tokens.

pfps:
I do not agree.
    https://plato.stanford.edu/entries/types-tokens/#WhaItNot
    
      Although the matter is discussed more fully in §8 below, it
        should
        be mentioned here at the outset that the type-token distinction
        is not
        the same distinction as that between a type and (what logicians
        call)
        its occurrences. Unfortunately, tokens are often
        explained as
        the “occurrences” of types, but not all occurrences of
        types are tokens. To see why, consider this time how many words
        there
        are in the Gertrude Stein line itself, the line type,
        not a
        token copy of it. Again, the correct answer is either three or
        ten, but
        this time it cannot be ten word tokens. The line is an
        abstract type with no unique spatio-temporal location and
        therefore
        cannot consist of particulars, of tokens. But as there are only
        three
        word types of which it might consist, what then are we counting
        ten of?
        The most apt answer is that (following logicians' usage) it is
        composed of ten occurrences of word types. See §8
        below,
        Occurrences, for more details.
    
    Further, type is used in RDF and ontologies as the relationship
      between an entity and classes that it belongs to (e.g., rdf:type)
      so it is better to avoid other possible meanings of type.
    
    
    peter
    PS:  I count 3 IRIs (actually 3 CURIES).  If I have to
      distinguish further, I count 5 occurrences of IRIs (CURIEs).  I
      count zero IRI (or CURIE) types and zero IRI (or CURIE) tokens.
    
    
    
    On 12/17/20 4:22 PM, Pierre-Antoine
      Champin wrote:
    
    
      
      
        Consider URIs. Are ex:a and ex:a two different tokens for the
          same type?
      
      Yes, if I read correctly https://plato.stanford.edu/entries/types-tokens/#WhaDis
        .
      
        Rose is a rose is a rose is a rose.

        In one sense of ‘word’ we may count three different words; in
          another sense we may count ten different words. C. S. Peirce
          (1931-58, sec. 4.537) called words in the first sense “types”
          and words in the second sense “tokens”.
      
      How many IRIs do you count in this?
      rdfs:Class rdf:type rdfs:Class;
           rdfs:subClassOf rdfs:Class.

      I count 3 IRI types, and 5 IRI tokens.
      I don't consider the terms of RDF (IRIs, literals...) and RDF*
        (+triples) to be different, in that respect, from the terms of
        the English language.
      Of course, we are not talking here about what the
        terms denote (as in the example in the link above: in
        "an 8,000 year old bean", does the word "bean" denote a bean
        type or a bean token?). We are talking about terms types and
        term tokens.
      —
        You are receiving this because you commented.
        Reply to this email directly, view it on GitHub, or unsubscribe.
      [
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/w3c/rdf-star/pull/75#issuecomment-747709535",
"url": "https://github.com/w3c/rdf-star/pull/75#issuecomment-747709535",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

pchampin:
>  type is used in RDF and ontologies as the relationship
      between an entity and classes that it belongs to (e.g., rdf:type)
      so it is better to avoid other possible meanings of type.

What about `peirce:type`, then? :wink:

More seriously: I was not advocating to use terms like "IRI type" and "IRI token" in the report -- as I didn't use them in my PR. I was only using Peirce's terminology in this discussion to show how the Type-Token distinction is, in my opinion, appropriate in this situation.

And yes, I also read the part about "what it is not" and occurrences. What I take away from it is: if you are counting IRIs in a graph or a triple, you are dealing with IRI occurrences. But if you are counting them on your screen or a printed page, then you are dealing with IRI tokens.

pfps:
I would call all occurrences of IRIs IRI occurrences, whether they are in a refresh of a screen (at 60 occurrences a second or more), on a screen, in a vocal utterance, in an email, in an email message, in a document, in a triple, or in a graph.  Tokens are used in too many other ways in computer science that could be used when talking about RDF and RDF*, so the "kind" of token needs to be distinguished so using the more generic occurrence ends up being less confusing.

As a case in point, programming languages have tokens (e.g., while) and these tokens have occurrences in code.

pfps:
The seminal example discussion is misguided.  Even if there is only stated-by information, and even if embedded triples are sensitive to syntax, a source states some occurrence of information.  The relationship between a source and a syntactic embedded triple is something like "stated an occurrence of" or "stated something that es expressed as".
Just state that this example is wrong and either stop there or give a good example with the needed indirection.

pchampin:
> The relationship between a source and a syntactic embedded triple is something like "stated an occurrence of" or "stated something that is expressed as"

Fully agreed. And I considered that `dct:source`'s definition was loose enough that one could interpret it like that.

> or give a good example with the needed indirection

There is a link to the example from section 2.1

pchampin:
This was discussed during today's call: https://w3c.github.io/rdf-star/Minutes/2020-12-18.html#item04

rat10:
> I would call all occurrences of IRIs IRI occurrences, whether they are in a refresh of a screen (at 60 occurrences a second or more), on a screen, in a vocal utterance, in an email, in an email message, in a document, in a triple, or in a graph. Tokens are used in too many other ways in computer science that could be used when talking about RDF and RDF*, so the "kind" of token needs to be distinguished so using the more generic occurrence ends up being less confusing.
> 
> As a case in point, programming languages have tokens (e.g., while) and these tokens have occurrences in code.

I just read up on the topic in SEP (https://plato.stanford.edu/entries/types-tokens/).  It seems indeed more appropriate to speak of 'types' vs 'occurrences' in our context as we are not concerned with renderings of triples on screens and printouts, but replacing 'token' by 'ooccurrence' everywhere seems extreme and not necessarily helpful.  Also we might like to differentiate between different serializations of the same occurrence (say in N-triples* vs Turtle*) by calling them 'tokens'. 
Then there is also the mismatch to the RDF 1.1 Semantics specification which talks about 'types' and 'tokens' but never uses the term 'occurrence', so we should explain our choice of terms.

rdf-star-cg-spec.html
```diff
@@ -1034,7 +1058,7 @@ <h2>Interpolation lemma</h2>
   <section class="appendix">
     <h2>Historical remarks</h2>
 
-    <section class="appendix">
+    <section>
       <h2>SA-mode and PG-mode</h2>
```
TallTed, line 1062:
```suggestion
      <h3>SA-mode and PG-mode</h3>
```

rdf-star-cg-spec.html
```diff
@@ -1047,6 +1071,29 @@ <h2>SA-mode and PG-mode</h2>
       </ul>
     </section>
 
+    <section>
+      <h2>The seminal example</h2>
```
TallTed, line 1075:
```suggestion
      <h3>The seminal example</h3>
```

TallTed (changes requested):
Falling within the `<h2>Historical remarks</h2>` as they do, these subsections should be demoted to `<h3>`

rdf-star-cg-spec.html
pchampin: From https://respec.org/docs/:
> Using h2 everywhere is sort of a tacit convention

plus, Respec changes the Hx tags itself into a consistent hierarchy.
Finally, even with H2 everywhere, this is still valid HTML...

TallTed: > > Using h2 everywhere is sort of a tacit convention

Ugh.  Respect is basically saying, "Be lazy and sloppy, authors!  We'll fix it in post!"

But I guess your dialect has now been blessed.

> plus, Respec changes the Hx tags itself into a consistent hierarchy.

*IFF* everything is properly wrapped and nested in `<section>` tags.  An important distinction without which the auto-hierarchy-construction will fail.

> Finally, even with H2 everywhere, this is still valid HTML...

I never said it wasn't valid HTML.  Plenty of valid HTML leads to incomprehensible browser rendering.  I'm glad that doesn't happen here.  I still believe it's best to keep the H# tags in the source numbered as they're intended to be rendered.

pchampin: > Ugh. Respect is basically saying, "Be lazy and sloppy, authors! We'll fix it in post!"

I look at it more positively, as "be careful to nest your sections correctly, and I'll take care of the tedious redundant numbering of H's" ;-)

> IFF everything is properly wrapped and nested in <section> tags. 

Of course, but my personal opinion is that <section>`s aren't optional nowadays. I think HTML would have been better off from the start with a `<section>` tag and a single `<h>` tag ;-)

TallTed: @pchampin - I cannot fully decipher your comment above, and neither can Github (as I cannot "quote reply" at all).  Can you correct the markdown therein, so it renders as intended?

TallTed:
@pchampin -- Note that there are currently conflicts that must be resolved before this can be merged, and the resolution of those conflicts may change the ongoing discussion as they will change the reading of the PR's results.

rdf-star-cg-spec.html
pchampin: Sorry, sent the previous comment too hastily :confused: This is now fixed.

TallTed: @pchampin - I'm sure I'm failing to mentally fill in something that should be obvious, but I'm even more sure that the markup is not leading to rendering as intended. Perhaps something was eaten/deleted when you saved it. As it stands, I think there's an extra backtick preceding `sections aren't optional` and some further imbalance as you go, with impact following `start with a`.

pchampin: :fearful: sorry again... should be good now

TallTed: *shrugs*  I think we must agree to disagree.  I prefer to keep the source as close as possible to the desired outcome, partly because it limits confusion when others review that source... which is where I always start, because the changes are highlighted better here than in the Respec DIFF.  Back to focusing on the content instead of the markup.  :-)

rdf-star-cg-spec.html
```diff
@@ -231,6 +231,30 @@ <h2>Concepts and Abstract Syntax</h2>
...
+
+      <p>In some situations, however, it might be necessary to distinguish the <em>occurrences</em> of a triple in different graphs. Consider the following sentence: "The triple `&lt;http://example.org/s&gt; &lt;http://example.org/p&gt; &lt;http://example.org/o&gt;` was added in (the graph represented by) file1.ttl by Alice, and the same triple was added in file2.ttl by Bob." Note that the words "same triple" in this sentence may be confusing, because although the triple (as an abstract entity) is the same, its respective occurrences are different things, each within a different file and with a different author (this is known, in philosophy and linguistics, as the <a href="https://en.wikipedia.org/wiki/Type%E2%80%93token_distinction">type-token distinction</a>). As the embedded triple represents a unique abstract entity, adequately conveying the meaning of the sentence above requires additional nodes for representing the distinct occurrences. One possible solution is illustrated in the following example (using the Turtle* concrete syntax described in <a href="#turtle-star">the next section</a>).</p>
```
hartig, line 240:
```suggestion
      <p>In some situations, however, it might be necessary to distinguish the <em>occurrences</em> of a triple in different graphs. Consider the following sentence: "The triple `&lt;http://example.org/s&gt; &lt;http://example.org/p&gt; &lt;http://example.org/o&gt;` in (the graph represented by) file1.ttl was added by Alice, and the same triple in file2.ttl was added by Bob." Note that the words "same triple" in this sentence may be confusing, because although the triple (as an abstract entity) is the same, its respective occurrences are different things, each within a different file and with a different author (this is known, in philosophy and linguistics, as the <a href="https://en.wikipedia.org/wiki/Type%E2%80%93token_distinction">type-token distinction</a>). As the embedded triple represents a unique abstract entity, adequately conveying the meaning of the sentence above requires additional nodes for representing the distinct occurrences. One possible solution is illustrated in the following example (using the Turtle* concrete syntax described in <a href="#turtle-star">the next section</a>).</p>
```

hartig: I think the given example sentence is somewhat misleading. By the way it is phrased, it is not entirely clear that this sentence is meant to about the occurrence of the triple rather than about the triple itself. In other words, when I read the sentence, I think: "Ah! Alice had this triple and she added it to the graph/file; additionally, Bob also had this triple and he added it to the other graph/file." So, based on this reading, it seems to me that the sentence is more about the triple itself rather than about its different occurrences. Now, my suggested change of the example sentence should fix this issue because it talks much more explicitly about the occurrences.

pfps:
There appears to be three changes in PR #75: a new section on triple occurrences,  a new section on the seminal example, and the moving of the original RDF* paper from an informative to a normative reference.  I find all three of these changes problematic.

Does the RDF* specification depend on anything in the original RDF* paper?  I heard not.  So the original RDF* paper should not be normative.

Is the seminal example problematic?  Yes, very!  So it should be disavowed.  The section on it does not and instead should read something like "The seminal example was wrong and should be completely disregarded."

What is an embedded triple?  Is it something like a literal or something like an IRI?  The new section is firmly on the side of literal.  This has consequences.  My view is that an embedded triple does  not need to be something like a literal.

pchampin:
@pfps
> So the original RDF* paper should not be normative.

Respec put it there, because of a mistake I made: I thought that appendices were automatically marked as "informative", but this is not the case. I marked the corresponding appendix as "informative", and the reference is back in the correct place. Thanks for spotting this.

For the other issues, I think it better to discuss them during the call.

rat10:
I think section 2.1 is a step in the right direction and the main load of appendix A.2 should be incorporated in section 2.1. Like:
- explain the problem and a possible workaround in 2.1 with the seminal example from Appendix A.2. 
- reduce the appendix to a mere mention of the historic development and a clarification that RDF* is now not the same as it was then, and until recently. 

It’s more important that people get it right now and in the future and that they are warned about a subtle but important change than that they understand what the historic background is. 

Appendix A.2 doesn’t explain what motivated the change. The explanations I could come up with are all not very flattering (like "at all cost avoid the semantic muddle that named graphs represent" or "we are so enarmored with the Superman problem" or "representing unasserted assertions suddenly seemed so tremedously important"). Of course if someone could put it a little more positively it would be good to add such an explanation about the motivation for the change as well.

pfps:
@rat10 Which change to RDF* affects the seminal example?

rat10:
> @rat10 Which change to RDF* affects the seminal example?

There was a discrepance between the seminal example referring to an occurrence and RDF* not saying how it addresses that occurrence. For a long time I considered this an oversight, sloppy engineering and/or something that people need to be educated about. The change was when a few weeks ago it became clear that RDF* will be specified in a way that makes the seminal example wrong, a regrettable mistake, whatever. Technically this is not a change in RDF* but just in the examples. Practically it is a change , and very much so IMO, which is why I find it important that section 2.1 addresses the problem rather comprehensively.

pfps:
Here is something that is closer to what is needed for A.2.  I would prefer something stronger and shorter but if there is a desperate need for a longer section on the seminal example, this wording at least lays out the situation more clearly.


A.2 The seminal example

The motivating example in the original RDF* paper [ RDF-STAR-FOUNDATION ] was on a provenance use-case, and is repeated below.

Example 19

:bob foaf:name "Bob".
<<:bob foaf:age 23>> dct:creator <http://example.com/crawlers#c1> ;
  dct:source <http://example.net/listing.html> .

This example is incorrect because there is a need to have multiple creators and related sources for the same embedded triple.  There is only one entity for an embedded in triple in RDF* so the source corresponding to a creator cannot be distinguished if provenance is represented in this manner.  Because of this, the example has given rise to significant confusion.

To rescue the example requires an intermediate entity to represent the stating of a triple, as in

<<:bob foaf:age 23>> ex:stating
  [ dct:creator <http://example.com/crawlers#c1> ;
  dct:source <http://example.net/listing.html> ] .

This corrected example shows that embedded triples can require more complex solutions than using RDF reification directly.

pfps:
@rat10  I agree that the seminal example only works if there are multiple entities for an embedded triple, but the semantics in the document had a single entity for an embedded triple and the newer semantics have also worked this way.  So no change to the meaning of RDF*, just an incorrect and misleading example.  (Well, using "just" here is really downplaying the seriousness of the situation.  Misleading examples can do vast amounts of damage.)

TallTed:
[I remain concerned](https://github.com/w3c/rdf-star/pull/75#issuecomment-756809167) that the merge conflicts on this PR may be concealing text that ought to be highlighted during our review and consideration of this PR.  Please, can those be addressed soon?

pchampin:
This was discussed during today's call https://w3c.github.io/rdf-star/Minutes/2021-01-15.html#item02

pchampin:
@TallTed conflict resolved

hartig:
Thanks for the input @pfps and @rat10, and apologies for not participating in yesterday's call (it was my daughter's birthday and I couldn't sneak out of the activities).

Some responses to what you write above and what was mentioned in some emails on the list:

I concur with [Peter's comment](https://github.com/w3c/rdf-star/pull/75#issuecomment-761015097) that there is no change from the original paper to our current draft in terms of the meaning of RDF* and, instead, it is only the original example that has been wrong and misleading. I appreciated your continued input to make sure we get the examples in the draft right.

I realize now that there are actually two separate mistakes that I made when writing this example.

One of them is related to the distinction between the notion of an RDF* triple as a single entity (whose identity is defined entirely by the three RDF* terms it consists of) and the notion of an occurrence of such a triple. I simply didn't make this distinction. I think that the new Section 2.1 in the current version of this PR here does a good job raising the readers' awareness of the need to make this distinction.

As a side node related to this mistake, I actually don't think that anywhere in the original paper the examples *explicitly* refer to occurrences of triples; I mean, nowhere in the text of the paper do I say that the example is about providing metadata for a specific occurrence of the triple about Bob's age. However, I can see now that readers may interpret the data in the example in this way, and that's one of the reasons the example is badly chosen as I understand now (and another reason is that I should have made the distinction as mentioned above).

The other mistake I made in the example is that the modeling of the metadata is insufficient. Instead of using the embedded triple about Bob's age directly as the subject of the two metadata triples, I should have introduced a separate entity that captures the creation of this triple (or, more precisely, as I know now, the creation of an occurrence of the triple) and then represent the information about this creation (i.e., creator and source) as triples with this separate entity as their subject.

A side node related to this mistake: I realize now that the given metadata triple with predicate `dct:source` may be interpreted to state where the embedded triple occurs. The current version of Section A.2 in this PR seems to say that this interpretation may have been implied by the example. However, that was not the intention of having this metadata triple in the example. In contrast, the intention was that this metadata triple specifies the HTML file that was the source used for the creation by the crawler.

Now, to address these mistakes, the RDF* graph of the example can be changed as follows (written in Turtle*, prefix declarations omitted).
```
_:a  :occurrenceOf  <<:bob foaf:age 23>> .
_:a  :creation  _:b .
_:b  dct:creator  <http://example.com/crawlers#c1> .
_:b  dct:source  <http://example.net/listing.html> .
```
Notice that the blank node labeled as `_:a` represents the occurrence of the triple about Bob's age, and the blank node labeled as `_:b` represents the entity that captures the creation of that triple occurrence.

As a final remark here, after these fixes, I would say that this example is not suitable anymore as the main *first* example to present the basic idea of RDF*. On the other hand, I think it is suitable now as an example that demonstrates that RDF* can be used as a building block to capture provenance use cases.

hartig:
@pfps in your comment above you write:

> To rescue the example requires an intermediate entity to represent the stating of a triple, as in
> 
> <<:bob foaf:age 23>> ex:stating
>  [ dct:creator http://example.com/crawlers#c1 ;
>  dct:source http://example.net/listing.html ] .
> 
>  This corrected example shows that embedded triples can require more complex solutions than using RDF reification directly.

While I am fine with the corrected example (it is a variation of the fix in my previous comment but without capturing the creation as a separate entity), I don't understand your remark about the "more complex solutions."

What you write in your corrected example are essentially three triples and one of them contains an embedded triple. So, it's four triples overall if we count the embedded triple as an independently. Let's compare this to RDF reification. While I am not entirely sure what you mean by "using RDF reification directly," I assume you mean to represent the example as follows (correct me if I am wrong).
```
_:x  rdf:type  rdf:Statement .
_:x  rdf:subject  :bob .
_:x  rdf:predicate  foaf:age .
_:x  rdf:object  23 .
_:x  dct:creator <http://example.com/crawlers#c1> .
_:x  dct:source <http://example.net/listing.html> .
```
In this representation, I am counting six triples.

Can you explain what you mean by "more complex"?

pfps:
RDF reification has statements, which are underspecified.  If one desires, one can use statements as, essentially, triple occurrences.  No extra kinds of entities are required.
RDF* has embedded triples.  In many cases there is also a need for triple occurrences.  So two new kinds of entities. 
As two is greater than one, I view the solution in RDF* as more complex.

hartig:
Interesting perspective! Without you mentioning it, I may never have looked at it this way. Probably that is because the number of entities is typically not so relevant for systems (in contrast to the number of triples that need to be stored and processed).

rdf-star-cg-spec.html
```diff
@@ -1041,9 +1065,32 @@ <h2>SA-mode and PG-mode</h2>
...
+
+      <pre data-transform="updateExample"
+        data-content-type="text/x-turtle-star"
+        class="nohighlight example"
+      >
+        <!--
+        # the controversial seminal example
+        :bob foaf:name "Bob".
+        <<:bob foaf:age 23>> dct:creator <http://example.com/crawlers#c1> ;
+                             dct:source <http://example.net/listing.html> .
```
ericprud, line 1081: I think the second triple should be written in its entirety so that readers don't think that the single occurrence of the *string* `<<:bob foaf:age 23>>` is a factor in determining that the two arcs discuss the same RDF* triple.

pchampin: @ericprud since this is a verbatim of the example in the original paper, I would rather keep it as is.

pchampin:
@pfps 
> What is an embedded triple? Is it something like a literal or something like an IRI? The new section is firmly on the side of literal. This has consequences. My view is that an embedded triple does not need to be something like a literal.

As you know, my position is that it is something like a literal. But I agree with you that this section does not need to be "tainted" by this assumption. In the latest commit (d0d948f), I slightly rephrased §2.1 to make it neutral w.r.t. this question.

pchampin:
@pfps 
>  As two is greater than one, I view the solution in RDF* as more complex.

The way I see it, the solutions in RDF* are less complex, because the language is more expressive We trade additional complexity in the language itself (adding an extra kind of terms) for less complexity in the graphs using that language (less triples) and the serializations encoding these graphs (thanks to `<<..>>` and `{|...|}`).

rat10:
> @pfps
> 
> > As two is greater than one, I view the solution in RDF* as more complex.
> 
> The way I see it, the solutions in RDF* are less complex, because the language is more expressive We trade additional complexity in the language itself (adding an extra kind of terms) for less complexity in the graphs using that language (less triples) and the serializations encoding these graphs (thanks to `<<..>>` and `{|...|}`).

You are always quick with adding a new indirection (or a new entity, as @pfps calls it) in your examples but I find those indirections quite problematic, especially as RDF provides so little means to tell where one complex object (composed of possibly many such entities) ends and the next begins. More triples with the same subject are generally easier to comprehend and query for than deeply nested structures.

Issue #37, opened 2020-11-19
Should RDF* be just syntactic sugar on top of RDF?


pchampin:
In other words, does RDF* need its own abstract syntax and semantics, or can it be "encoded" in standard RDF?

It largely depends on the answer to issue #22.
1. If we want embedded triples to be referentially transparent, then they can be internally represented using, e.g., standard reification or singleton properties.
2. If we want them to be referentially opaque, and forbid them to contain blank nodes, then they can be internally represented using specific IRIs (see #23).
3. Otherwise, IMO, we need to somehow extend RDF semantics. However, I see still see two paths here:
  a) either we promote RDF* triples as a new kind of terms, as done in the original papers and the current version of the report, or
  b) we extend RDF's semantics with a built-in datatype for *representing* IRIs and literals, and we represent RDF* triples using an adapted form of reification.

To illustrate the last bullet:
```
<< :alice :age 26 >> :accordingTo :bob.
```
could be seen as syntactic sugar for
```
_:stmt :accordingTo :bob.
_:stmt
    rdfx:subjectTerm "<http://example.org/alice>"^^rdfx:term;
    rdfx:predicateTerm "<http://example.org/age>"^^rdfx:term;
    rdfx:subjectTerm "\"26\"^^<www.w3.org/2001/XMLSchema#integer>"^^rdfx:term.
```
There are several reasons why I believe this modelling needs a small extension to RDF semantics, but I'll develop them if we come to a point where we consider this option seriously...

pchampin:
DISCLAIMER: if we decided that RDF* can be encoded in standard RDF (e.g. using standard reification), that would not impose that implementers store it that way in their systems. They could still use internal optimizations, e.g. to avoid the overhead of storing additional 3 or 4 triples for every reified triple...

afs:
Good issue. If we can decide this, we can move on.

On the idea of encoding in triples -- using multiple triples, and a well-formed-ness condition -- as a basic level feature:

Long ago, Jena had special code to optimize reification in storage. Jena does not have such code any longer - it was virtually unused and created a significant cost on building storage subsystems.

The trouble is not reification itself, it is the well-formed-ness condition and handling of partial reification, unfinished reification and wrong reification.

[...]

PR #81, opened 2021-01-07
Proposal for a syntactic-sugar based semantics

pchampin:
Related to #37

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/pull/81.html" title="Last updated on Feb 4, 2021, 5:44 PM UTC (4943713)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/81/e2fc2f6...4943713.html" title="Last updated on Feb 4, 2021, 5:44 PM UTC (4943713)">Diff</a>

### DIFF
--- a/rdf-star-cg-spec.html
+++ b/rdf-star-cg-spec.html
@@ -150,7 +150,7 @@ SELECT ?claimer WHERE {
 }
 -->
 </pre>
-    
+
 
 <p> SPARQL query triple patterns that include a triple pattern as a subject or object are known as SPARQL* triple patterns. </p>
 
@@ -351,7 +351,7 @@ SELECT ?claimer WHERE {
 
     <section>
       <h2>Parsing</h2>
-      <p>A Turtle* parser is similar to a Turtle parser as defined in <a data-cite="TURTLE#h2_sec-parsing">Section 7 of the Turtle specification</a> [[TURTLE]], with an additional item in its state :</p>
+      <p>A Turtle* parser is similar to a Turtle parser as defined in <a data-cite="TURTLE#h2_sec-parsing">Section 7 of the Turtle specification</a> [[TURTLE]], with an additional item in its state:</p>
       <ul>
         <li id="curObject"><a>RDF* Term</a> |curObject| — The |curObject| is bound to the <a href="#grammar-production-embObject">`embObject`</a> production.</li>
       </ul>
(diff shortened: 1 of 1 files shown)


rdf-star-cg-spec.html
```diff
@@ -801,222 +801,115 @@ <h2>SPARQL* Update</h2>
...
-            <li>false otherwise</li>
-          </ul>
-        </ul>
-
-        <p>Since |IL| and |IT| are partial mappings, |I|(|E|) may be undefined for some <a>literal</a> or <a>triple</a> |E|. In that case, |E| has no semantic value in |I|, so any <a>asserted triple</a> having E as <a>subject</a> or <a>object</a> it will fail to satisfy the condition above, hence any <a>graph</a> containing such <a>asserted triple</a> will be false.</p>
+      <p>We introduce a new <a>datatype</a>, denoted by the <a>IRI</a> <var>D*</var>, and defined as follow:</p>
+      <ul>
+        <li>its <a>value space</a> is the union of the set of <a>IRIs</a> with the set of <a>Literals</a>.</li>
+        <li>its <a>lexical space</a> is the set of strings satisfying the <a data-cite="N-TRIPLES#grammar-production-IRIREF">`IRIREF`</a> or <a data-cite="N-TRIPLES#grammar-production-literal">`literal`</a> production rules of the N-Triples grammar [[N-TRIPLES]], furthermore complying with the <a data-cite="N-TRIPLES#h2_canonical-ntriples">canonical form</a>, and having either an explicit datatype or language tag for `literal` (i.e. no <a>simple literal</a>).</li>
```
TallTed, line 829:
```suggestion
        <li>its <a>lexical space</a> is the set of strings satisfying the <a data-cite="N-TRIPLES#grammar-production-IRIREF">`IRIREF`</a> or <a data-cite="N-TRIPLES#grammar-production-literal">`literal`</a> production rules of the N-Triples grammar [[N-TRIPLES]], furthermore complying with the <a data-cite="N-TRIPLES#h2_canonical-ntriples">canonical form</a>, and having either an explicit datatype or language tag for `literal` (i.e., no <a>simple literal</a>).</li>
```

rdf-star-cg-spec.html
```diff
@@ -801,222 +801,115 @@ <h2>SPARQL* Update</h2>
...
+        <li>its <a>lexical-to-value mapping</a> consists in parsing the <a>lexical form</a> as specified by [<a data-cite="N-TRIPLES#h3_sec-parsing-terms">N-TRIPLES, Section 8.1</a>].</li>
+      </ul>
+      <p>Any <a>RDF* interpretation</a> (see below) will be required to <a>recognize</a> this <a>datatype</a>, i.e. to correctly map <a>literals</a> of that datatype to the appropriate <a>IRI</a> or <a>literal</a>.</p>
```
TallTed, line 832:
```suggestion
      <p>Any <a>RDF* interpretation</a> (see below) will be required to <a>recognize</a> this <a>datatype</a>, i.e., to correctly map <a>literals</a> of that datatype to the appropriate <a>IRI</a> or <a>literal</a>.</p>
```

rdf-star-cg-spec.html
```diff
@@ -801,222 +801,115 @@ <h2>SPARQL* Update</h2>
...
+      <p>Given an <a>RDF* graph</a> |G|, the following steps transform it into an <a>RDF graph</a> that we call |unstar|(|G|).</p>
 
-        <ul>
-          <li>if |E| is a <a>blank node</a> then <var>Γ</var>*(|E|) = <var>Γ</var>(|E|) if it is defined, |E| otherwise</li>
-          <li>if |E| is an <a>IRI</a> or a <a>literal</a>, then <var>Γ</var>*(|E|) = |E|</li>
-          <li>if |E| is an <a>RDF* triple</a> (|s|,|p|,|o|), then <var>Γ</var>*(|E|) = (<var>Γ</var>*(|s|),<var>Γ</var>*(|p|),<var>Γ</var>*(|o|))</li>
-          <li>if |E| is an <a>RDF* graph</a>, then <var>Γ</var>*(|E|) = {<var>Γ</var>*(|t|) for all |t| ∈ |E|} </li>
-        </ul>
+      <ol>
+        <li>Define an injective mapping |Q| from the set of <a>mixed blank nodes</a> of |G| to a set of fresh (i.e. not occurring in |G|) <a>blank nodes</a>.</li>
```
TallTed, line 843:
```suggestion
        <li>Define an injective mapping |Q| from the set of <a>mixed blank nodes</a> of |G| to a set of fresh (i.e., not occurring in |G|) <a>blank nodes</a>.</li>
```

rdf-star-cg-spec.html
```diff
@@ -801,222 +801,115 @@ <h2>SPARQL* Update</h2>
...
+        <li>Extend |Q| to other <a>RDF* terms</a> as follow:<ul>
```
TallTed, line 844:
```suggestion
        <li>Extend |Q| to other <a>RDF* terms</a> as follows:<ul>
```

rdf-star-cg-spec.html
```diff
@@ -801,222 +801,115 @@ <h2>SPARQL* Update</h2>
...
+    <section>
+      <h2>RDF* Entailment</h2>
 
-        <p>Any <a data-cite="RDF11-MT#dfn-semantic-extension">semantic extension</a> of RDF MAY be extended to RDF* by replacing the semantic conditions, the notion of satisfiability and the notion of entailment, defined in [[[RDF11-MT]]], by their corresponding extension defined above. This is notably the case for <a data-cite="RDF11-MT#D_entailment">Datatype entailment</a> and <a data-cite="RDF11-MT#rdfs_entailment">RDFS entailment</a>.</p>
-      </section>
+      <p>Following [[[RDF11-MT]]], we extend the notions of satisfiability and entailment for <a>RDF* interpretations</a>. Given a set |D| of <a>IRIs</a> denoting <a>datatypes</a> and two <a>RDF* graphs</a> |G| and |H|:<p>
+      <ul>
+        <li>An <a>RDF* interpretation</a> |I| <dfn data-lt="satisfying|satisfies">satisfies</dfn> |G| if |I|(|unstar|(|G|)) = true. In the particular case of <a>RDF graphs</a>, this is equivalent to |I|(|G|) = true, consistently with the definitions in [[RDF11-MT]].</li>
```
TallTed, line 877:
```suggestion
        <li>An <a>RDF* interpretation</a> |I| <dfn data-lt="satisfying|satisfies">satisfies</dfn> |G| if |I|(|unstar|(|G|)) = true. In the particular case of <a>RDF graphs</a>, this is equivalent to |I|(|G|) = true, consistent with the definitions in [[RDF11-MT]].</li>
```

rdf-star-cg-spec.html
```diff
@@ -801,222 +801,115 @@ <h2>SPARQL* Update</h2>
...
-          <!--
-          SELECT ?who { :alice foaf:knows ?who }
-          -->
-        </pre>
-      </section>
-
-      <section id="referential-opacity">
-        <h2>Referential opacity</h2>
+      <h2>Remarks</h2>
+      <p>Any <a>ground</a> <a>embedded triple</a> (|s|, |p|, |o|) is technically replaced by a blank node, acting as an existential variable, but it is pinpointed by <var>S*</var>, <var>P*</var> and <var>O*</var> to a single element of the universe, which is the triple (|s|, |p|, |o|) itself. As such, it behaves similarly to a literal (it denotes the same thing in every interpretation).</p>
```
TallTed, line 886:
```suggestion
      <p>Any <a>ground</a> <a>embedded triple</a> (|s|, |p|, |o|) is technically replaced by a blank node, acting as an existential variable, but it is pinpointed by <var>S*</var>, <var>P*</var> and <var>O*</var> to a single element of the universe, which is the triple (|s|, |p|, |o|) itself. As such, it behaves similarly to a literal (i.e., it denotes the same thing in every interpretation).</p>
```

rdf-star-cg-spec.html
```diff
@@ -801,222 +801,115 @@ <h2>SPARQL* Update</h2>
...
 
-        <p class="issue">We didn't prove it yet...</p>
-      </section>
+      <p>Finally, it is worth noticing that the <a>IRIs</a> <var>D*</var>, <var>S*</var>, <var>P*</var>, <var>O*</var> and <var>I*</var> are not explicitly specified (besides their existence) by this document. With this omission, this specification provides no standard way to convey an <a>RDF* graph</a> using older RDF concrete syntaxes – only the RDF*-specific extensions of those concrete syntaxes can be used for that. Otherwise, it would have been possible to describe incomplete or ill-formed <a>RDF* triples</a> (a problem that exists with <a data-cite="RDF-SCHEMA#ch_collectionvocab">RDF collections</a> for example), thus forcing implementations to deal with pathological cases, and preventing useful optimisations.</p>
```
TallTed, line 908:
```suggestion
      <p>Finally, it is worth noticing that the <a>IRIs</a> <var>D*</var>, <var>S*</var>, <var>P*</var>, <var>O*</var> and <var>I*</var> are not explicitly specified (beyond their existence) by this document. With this omission, this specification provides no standard way to convey an <a>RDF* graph</a> using older RDF concrete syntaxes &mdash; only the RDF*-specific extensions of those concrete syntaxes can be used for that. Otherwise, it would have been possible to describe incomplete or ill-formed <a>RDF* triples</a> (a problem that exists with <a data-cite="RDF-SCHEMA#ch_collectionvocab">RDF collections</a> for example), thus forcing implementations to deal with pathological cases, and preventing useful optimisations.</p>
```
I think this paragraph needs further work, as I do not understand what's meant by the text following the first sentence, and more, how/why what's said in those latter sentences follow from what's said in the first.

pchampin: Imagine that the spec specified S* = `rdfx:s`, P* = `rdfx:p` ... and defined a specific IRI for prefix `rdfx:`. Then, the following Turtle(*) document:
```
[]  rdfx:s "<http://example.org/s1>^^rdfx:d;
    rdfx:p "<http://example.org/p1>^^rdfx:d;
    rdfx:o "<http://example.org/o1>^^rdfx:d;  
    dct:creator :alice.
```
would be semantically equivalent to the following Turtle* document:
```
<<:s1 :p1 :o1>> dct:creator :alice.
```
and should therefore be parsed as such by compliant RDF* systems.

More tricky, the following Turtle(*) document:
```
[] rdfx:s "<http://example.org/s1>^^rdfx:d;
   dct:creator :alice.
```
would be semantically equivalent to the following Turtle* document:
```
<<:s1 [] [] >> dct:creator :alice.
```
and should therefore be parsed as such by compliant RDF* systems.

The following Turtle(*) document:
```
[] rdfx:s "<http://example.org/s1>^^rdfx:d, "<http://example.org/s2>^^rdfx:d.
```
would be inconsistent, and that also would have to be dealt with.

To sum it up, that would cause too much burden on RDF* systems. So instead of opening the pandora box, we define the mapping on a theoretical level only.

Does that make more sense? If so, how would you suggest to rephrase the above paragraph?

pchampin:
This was discussed during today's call https://w3c.github.io/rdf-star/Minutes/2021-01-15.html#item03

rdf-star-cg-spec.html
```diff
@@ -861,12 +861,18 @@ <h2>RDF* Interpretation</h2>
...
+
+        <p>It is important to understand the <a href="https://en.wikipedia.org/wiki/Use%E2%80%93mention_distinction">use-mention distinction</a> at play here. <a>RDF* terms</a> <em>used</em> in the abstract syntax denote any element of |IR|, while <a>RDF* terms</a> belonging to |IR| may be denoted (or <em>mentioned</em>) by a different term, regardless of their own denotation, if any. For example, the <a>IRI</a> `&lt;http://www.w3.org/1999/02/22-rdf-syntax-ns#nil&gt;` belongs to |IR| (all <a>IRIs</a> do), and can be mentioned by using the <a>literal</a> `"&lt;http://www.w3.org/1999/02/22-rdf-syntax-ns#nil&gt;"^^D*` (which denotes that <a>IRI</a> by definition of <var>D*</var>). On the other hand, when used in the graph, this <a>IRI</a> denotes an empty list.</p>
```
ericprud, line 873: should that quoted rdf:nil have a datatype xsd:anyURI?

pchampin: Yes, it could be argued that xsd:anyURI is a subclass of the proposed D* datatype. But between the fact that we need IRIs rather than URIs, and that we also need literals, defining a single datatype covering all the required terms seemed a simpler option.

pchampin:
PR #88 has been merged instead.

PR #83, opened 2021-01-08
More combinations on embedded syntax

gkellogg:
@afs I didn't see a script for generating manifest.jsonld from manifest.ttl; I updated sparql/syntax/manifest.jsonld by hand, but didn't update turtle/eval/manifest.jsonld. I did build the HTML versions, however.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -1,5 +1,6 @@
 .PHONY: all
 all:
+	#./make-jsonld-manifests.rb
 	./make-html-manifests.py
 
 .PHONY: clean
--- a/tests/make-html-manifests.py
+++ b/tests/make-html-manifests.py
@@ -9,7 +9,7 @@ DIR = Path(__file__).parent
 
 def main():
     for i in [
-        ["semantics", "manifest.jsonld"],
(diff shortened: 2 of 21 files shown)


afs (approved):

afs:
I don't have a script - the "semi-mechanical" process I used was to convert manifest.ttl to JSON-LD using jsonld-java (via Jena `riot --pretty jsonld`) then manually edit to make it nicer as JSON - inlining the tests into the entries array (separate objects otherwise) and making URIs relative.

From then on, it a matter of copying and new tests into manifest.json, with the fragment taken from a new run of `riot --pretty jsonld` - they does produce the right fragment apart from the absolute `@id`.

Then check they produce the same RDF and also run the same tests.

gkellogg:
Okay, I have a script that does that too. I’ll regenerate and push up.

pchampin (approved):
Looking good to me. Do you (@afs @gkellogg @hartig) think that we should discuss this PR on call time?

gkellogg:
I don’t think it warrants call time, it’s just adding coverage. The details for how the manifests are managed shouldn’t affect anyone.

PR #88, opened 2021-02-04
Alternative semantics based on @pfps' proposal

pchampin:
source: https://lists.w3.org/Archives/Public/public-rdf-star/2021Jan/0059.html

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/pull/88.html" title="Last updated on Feb 12, 2021, 6:37 PM UTC (2ad1ef3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/88/5c56a98...2ad1ef3.html" title="Last updated on Feb 12, 2021, 6:37 PM UTC (2ad1ef3)">Diff</a>

### DIFF
--- a/rdf-star-cg-spec.html
+++ b/rdf-star-cg-spec.html
@@ -919,7 +919,7 @@ SELECT ?claimer WHERE {
       and <dfn data-cite="RDF11-MT#dfn-equivalent" data-lt="equivalent">equivalence</dfn>.
     </p>
 
-    <section>
+    <section id="mapping">
       <h2>Mapping RDF* abstract syntax to RDF</h2>
 
       <p>We consider six <a>IRIs</a> |ST:subject term|, |PT:predicate term|, |OT:object term|, |SS:subject string|, |PS:predicate string| and |OS:object string| that will have a special meaning in our mapping.</p>
@@ -933,7 +933,7 @@ SELECT ?claimer WHERE {
       <p>Given an <a>RDF* graph</a> |G|, the following steps transform it into an <a>RDF graph</a> that we call |unstar|(|G|).</p>
 
       <ol>
-        <li>Pick an <a>RDF* triple</a> (|s|, |p|, |o|) in the <a>constituents</a> |G| such that neither |s|, |p| nor |o| is an <a>embedded triple</a>.</li>
+        <li>Pick an <a>RDF* triple</a> (|s|, |p|, |o|) in the <a>constituents</a> of |G| such that neither |s|, |p| nor |o| is an <a>embedded triple</a>.</li>
         <li>Mint a fresh blank node |b|, and replace by |b| all occurrences of (|s|, |p|, |o|) in the subject or object position of an <a>asserted</a> or <a>embedded triple</a> of |G|.</li>
         <li>Add the following <a>asserted triples</a> to |G|:<ul>
           <li>(|b|, |ST|, |s|)</li>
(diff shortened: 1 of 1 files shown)


rdf-star-cg-spec.html
```diff
@@ -829,9 +829,66 @@ <h2>SPARQL* Update</h2>
...
+      <p>We define a mapping |L| that maps any term |t| that is either an <a>IRI</a> or a <a>literal</a> to a literal with <ul>
+        <li>`xsd:string` as its the <a>datatype</a>, and</li>
+        <li>the <a data-cite="N-TRIPLES#canonical-ntriples">canonical N-Triples</a> representation of |t| as its <a>lexical form</a> [[N-TRIPLES]]. If `t` is itself a literal with the `xsd:string` <a>datatype</a>, the representation MUST NOT include the datatype.</li>
+      </ul>
+      </p>
+
+      <p>Given an <a>RDF* graph</a> |G|, the following steps transform it into an <a>RDF graph</a> that we call |unstar|(|G|).</p>
+
+      <ol>
+        <li>Pick an <a>RDF* triple</a> (|s|, |p|, |o|) in the <a>constituents</a> |G| such that neither |s|, |p| nor |o| is an <a>embedded triple</a>.</li>
```
TallTed, line 859:
```suggestion
        <li>Pick an <a>RDF* triple</a> (|s|, |p|, |o|) in the <a>constituent</a> |G| such that neither |s|, |p| nor |o| is an <a>embedded triple</a>.</li>
```

pchampin: Actually that should be "in the constituents of G". Thanks for spotting it. It's fixed in my local copy and will be fixed in the next commit.

pchampin:
@pfps you might want to look at this PR, see if I captured your ideas correctly.
That is, apart from the "hidden IRIs" part which I know you are no a big fan of... About that, I included some text in the "remarks" to (hopefully) better explain the rationale.

rdf-star-cg-spec.html
```diff
@@ -868,7 +959,7 @@ <h2>The seminal example</h2>
...
+      <p!>This example was further debated on the <a href="https://lists.w3.org/Archives/Public/public-rdf-star/">RDF* mailing list</a>, as it appears to have set wrong expectations about what RDF* embedded triples represent. More precisely, from this example, one may wrongly assume that `&lt;&lt;:bob foaf:age 23>>` represents the <em>occurrence</em> of the given triple at the address `http://example.net/listing.html` (see <a href="#occurrences"></a>). This impression may be reinforced by the use of `dct:creator`: arguably, a triple (as a unique abstract entity) is not "created" by anyone, while an occurrence thereof can be said to be created or authored.</p>
```
ericprud, line 962: is `<p!>` a typo or a processing instruction for some magic preprocessor i'm unaware of?

pchampin: the former I'm affraid!

pfps:
This PR appears to be similar to the semantics I stated.  The use of unnamed predicates is a significant difference.  My proposed used the standard RDF reification vocabulary and added particular predicate IRIs for a purpose - users can intermix RDF* syntax with RDF reification syntax.

rdf-star-cg-spec.html
```diff
@@ -829,9 +829,100 @@ <h2>SPARQL* Update</h2>
   <section>
     <h2>RDF* Semantics</h2>
 
-    <p>In this section, we provide a model-theoretic semantics for RDF*, by extending the one defined in [[[RDF11-MT]]] [[RDF11-MT]].</p>
+    <p>In this section, we provide a model-theoretic semantics for RDF*, based on the one defined in [[[RDF11-MT]]] [[RDF11-MT]]. More precisely, we define a mapping from RDF*'s abstract syntax into standard RDF's abstract syntax, and define the semantics of <a>RDF* graphs</a> in terms of the semantics of the mapped <a>RDF graphs</a>.</p>
```
pfps, line 832: In one place embedded triples can be predicates, but in one other place only embedded triples as subjects or objects are handled.

ericprud: @pchampin , i read this as a proposal to:
```
s/such that neither s, p nor o is an embedded triple.
 /such that neither s nor o is an embedded triple.
 /
```

pchampin: Right; the original proposal was very general, allowing any kind of term in any position. I took the liberty to make it more aligned with the restrictions in the abstract syntax, but I let this one go.

gatemezing:
Section 6.1: Would it be possible to suggest ST, PT, OT, SS, PS, and OS IRIs as per @pfps email? even if it is not normative? I'm also concerned about 6.3.2.  It reads like RDF-star graphs (in general) are not interoperable (modulo mapping between the different ST, PT, OT, etc). Would that affect adoption? that's another question.
We might also consider having this section in the document coming before SPARQL*, Turtle*.

hartig:
I find Sec.6.3.2 quite clear and convincing. Perhaps the beginning of of the second paragraph may be phrased in a more positive way.

So, my question in this context is actually the opposite: why should we introduce actual IRIs for ST, PT, OT, SS, PS, and OS? What would be the benefit?

rdf-star-cg-spec.html
```diff
@@ -866,7 +867,7 @@ <h2>Mapping RDF* abstract syntax to RDF</h2>
           <li>(|b|, |PS|, |L|(|p|))</li>
           <li>(|b|, |OS|, |L|(|o|)) unless |o| is <a>blank node</a></li>
         </ul></li>
-        <li>Repeats the steps above until there are no embedded triple left in |G|</li>
+        <li>Repeat the steps above until there are no embedded triple left in |G|.</li>
```
TallTed, line 870:
```suggestion
        <li>Repeat the steps above until there are no embedded triples left in |G|.</li>
```

hartig:
A question regarding the property described in Sec.6.3.1: This property seems undesirable to me because it may cause unexpected behavior. Can some explain why this would not be a problem in practice?

pfps:
It appears to me that 6.3.1 does not correctly describe the issue.  The issue is not merging RDF* graphs, but merging RDF graphs.  If implementations work with only with RDF* graphs then the issue should not arise.

rdf-star-cg-spec.html
```diff
@@ -829,9 +829,101 @@ <h2>SPARQL* Update</h2>
...
+    </section>
 
-    <div class="issue" data-number="81">one proposal on the table</div>
+    <section class="informative">
+      <h2>Remarks</h2>
+
+      <section>
+        <h2>Combining RDF* graphs</h2>
+
+        <p>Care must be taken when <a>RDF* graphs</a> are combined through union or <a>merging</a>. Given two <a>RDF* graphs</a> |G| and |H|, it may be the case that |unstar|(|G| ∪ |H|) ≠ |unstar|(|G|) ∪ |unstar|(|H|). More precisely, if |G| and |H| contain the same <a>embedded triple</a>, this triple will be mapped to a single blank node in |unstar|(|G| ∪ |H|), but in two potentially different blank nodes in |unstar|(|G|) ∪ |unstar|(|H|). These blank nodes will need to be unified in order to get the correct entailments.</p>
```
pchampin, line 893:
```suggestion
        <p>Care must be taken when <a>RDF* graphs</a> are combined through union or <a>merging</a> if those graphs have been mapped beforehand to <a>RDF graphs</a>. Given two <a>RDF* graphs</a> |G| and |H|, it may be the case that |unstar|(|G| ∪ |H|) ≠ |unstar|(|G|) ∪ |unstar|(|H|).More precisely, if |G| and |H| contain the same <a>embedded triple</a>, this triple will be mapped to a single blank node in |unstar|(|G| ∪ |H|), but in two potentially different blank nodes in |unstar|(|G|) ∪ |unstar|(|H|). These blank nodes will need to be unified in order to get the correct entailments.</p>
```
@pfps would that address [your comment](https://github.com/w3c/rdf-star/pull/88#issuecomment-778285079)

pchampin:
@pfps I am not sure whether you get notified for change suggestions, so just in case:
https://github.com/w3c/rdf-star/pull/88/files#r575393794
Would this fix address your comment above?

pfps:
I would say instead:

Care must be taken when RDF graphs that result from RDF* graphs are combined through union or <a>merging</a>.

PR #90, opened 2021-02-08
GitHub Actions CI

gkellogg:
This uses ReSpec to build and check the spec, including markup errors.

Note that it currently finds three issues with either directly, or indirectly linked resources. Some should be fixed. The link checking can be turned off.

It would run on every push and PR, as configured, and generally show a pass/fail as part of the PR.


kinow (approved):

pchampin:
Good idea, thanks.

PR #99, opened 2021-02-18
removes undesired background image from the published draft

hartig:
[As discussed on the mailing list](https://lists.w3.org/Archives/Public/public-rdf-star/2021Feb/0047.html), I think that the background picture with the light-red text saying "UNOFFICIAL DRAFT" should be removed from the published version of the draft. This PR does that.


HolgerKnublauch:
FWIW I did the same trick in other documents where I wanted to reuse the Respec document style. The watermarks can be quite annoying at times.

hartig:
@pchampin  ping

pchampin:
Done in c1dfc19 with a smaller change (and easier to apply on future versions).

PR #114, opened 2021-02-24
Basic SPARQL-star eval tests

afs:
SPARQL-star eval tests.

This is a snapshot of work-in-progress.


gkellogg:
We should add "eval/manifest.jsonld" to "sparql/manifest.jsonld" and decide if "manifest-query.jsonld" and "manifest-update.jsonld" still belong.

afs:
@gkellogg Yes (already done locally).

I'm only producing one manifests at the moment because it keeps changing; I'd like some advice on making JSON-LD for eval tests look nice when the suite is moderately stable.

gkellogg:
There’s a make task to do it, but it just uses a standard JSON serializer. Otherwise, you could write a customized one to control white space, but I think trying to maintain that may be more trouble than it’s worth.

gkellogg (approved):

afs:
This includes conversion "RDF*" to "RDF-star" in the Turtle and SPARQL test areas,  and a general test cleanup, as separate commits.

afs:
This is ready for implementers to try out:

It covers:
* result sets
* single triple pattern matches
* basic graph patterns
* CONSTRUCT
* GRAPH

Still more to do:
* [ ] Functions
* [ ] SPARQL Update
* [ ] Fix HTML generation for eval tests
* [ ] Acceptable JSON-LD

gkellogg (changes requested):
We need a sparql/eval/manifest.jsonld (and .html?). And sparql/manifest.jsonld should probably reference that version.

There's a Make task that will create both.

afs:
> We need a sparql/eval/manifest.jsonld 

Yes - this is on the todo list above "Acceptable JSON-LD". 

Previous I have crafted a template that makes JSON that uses  "tests/manifest-context.jsonld" so is more "JSON" and less "RDF". Could make-jsonld-manifests use that? Currently, it will overwrite the existing files. 

I just haven't figured out how to make it work with the SPARQL eval requirements.

> (and .html?)

make-html-manifests.py does not work - my guess is that it is because the actions are bnodes of a pair or more triples.

pchampin:
> make-html-manifests.py does not work - my guess is that it is because the actions are bnodes of a pair or more triples.

that's correct. I need to fix it...

afs:
I just haven't had time so it was either delay and remain in draft or put it for review now. The status is written in the comment above.

It going to cover several areas. I thought it was more useful to have this first part now.

tests/sparql/eval/data-7.ttl
```diff
@@ -0,0 +1,18 @@
+PREFIX : <http://example/>
+
+# One each of each collection class.
```
TallTed, line 3:
```suggestion
# One of each collection class.
```

tests/sparql/eval/manifest.html
```diff
@@ -1448,6 +1449,72 @@ <h2>SPARQL-star - Embedded triple - Functions <a href="#sparql-star-expr-2">🔗
...
+
+SELECT * {
+   :item :p ?v .
+} ORDER BY ?v
+</pre>
+<div>Data: <code><a href="data-7.ttl">data-7.ttl</a></code></div>
+<pre>
+PREFIX : &lt;http://example/&gt;
+
+# One each of each collection class.
```
TallTed, line 1470:
```suggestion
# One of each collection class.
```

tests/sparql/eval/data-7.ttl
afs: Doesn't read well but it is as meant.  It's not "one of the classes", its one instance of each class.

"One item of each of the collation classes." is clearer?

TallTed: My sentence does not say "one of the classes"; it effectively says "one [item] from each collection class."  (The word "item" is unnecessary, but can be included.)

"We have apples, oranges, and bananas. Which would you like?"
"One of each [collection class]."
Means you get one apple, one orange, and one banana ... which I believe is your intent.

afs: This is test data.
Changed to:
```
# The different kinds of orderable items.
# (Not in collation order).
```

gkellogg (approved):

gkellogg:
Looks good, they're all passing for me now.

We need to make sure the grammar is updated to add EmbTP (or whatever it's called now) to PrimaryExpression.

hartig:
> We need to make sure the grammar is updated to add EmbTP (or whatever it's called now) to PrimaryExpression.

It's on my agenda to work on the PR for that.

afs:
Rebased to current main.

Issue #101, opened 2021-02-19
hidden predicates can't hide from SPARQL


pfps:
Even if the reification predicates are given IRIs that cannot be written in RDF* or RDF documents, as in https://w3c.github.io/rdf-star/cg-spec/2021-02-18.html#mapping, the following SPARQL query can be used to retrieve such triples:

```
SELECT ?s ?p ?o WHERE { ?s ?p ?o }
```

This appears to affect all semantics for RDF* that use hidden predicates.

afs:
I believe that it would be better to define SPARQL-star on RDF-star datasets.

https://www.w3.org/TR/sparql11-query/#BasicGraphPattern

has the way matching is BGP's is performed.

The extension entailment framework does need attention.

https://www.w3.org/TR/sparql11-query/#sparqlBGPExtend

pfps:
Isn't SPARQL defined in term of simple entailment?   It seems to me that that requires a full semantics for RDF*.

afs:
(as discussed in the telecon today):

SPARQL matching:
https://www.w3.org/TR/sparql11-query/#defn_PatternInstanceMapping

Just simple entailment results in infinite results - SPARQL needs to retain (e.g.) which blank node matched so the algebra (e.g. OPTIONAL) works.

Any entailment regimes matching has some specific requirements:
https://www.w3.org/TR/sparql11-query/#sparqlBGPExtend

pfps:
I spent some time trying to find out how to define querying in terms of entailment.  It's not easy.

pfps:
It turns out that the support for querying as simple entailment easily comes from the RDF 1.1 Semantics document.  I was trying to show the result from first principles but this is not necessary.

In Appendix C there is the following proof:

Interpolation Lemma (from Section 5.3):
G simply entails a graph E if and only if a subgraph of G is an instance of E.

Proof (from Appendix C):
[...]

PR #127, opened 2021-03-05
New semantics proposal, using "lower & upper bounds"

pchampin:
Here is an adaptation of the semantics, that I believe has some advantages over the current version:

* ~~it does not rely anymore on "hidden" predicates (#101)~~
* it does not have the "merging" issue warned about in [§6.3.1](https://w3c.github.io/rdf-star/cg-spec/2021-02-18.html#combining-rdf-star-graphs)
* I think that it allows us to align SPARQL query semantics with simple entailment (as newly defined)
* I think that it allows the [Interpolation Lemma](https://www.w3.org/TR/rdf11-mt/#dfn-interpolation) to extend to RDF-star
* it still uses one "hidden" IRI, but this once can hide from SPARQL (#101)

(I didn't formally prove the last two items, hence "I think"...)

The trick is that we do not map anymore RDF-star graphs to a single, semantically equivalent RDF graph.
Instead, we map it to a pair of RDF graphs, which can be thought of as a "lower and upper bound" of the RDF-star graph, in terms of entailment. The semantics of the RDF-star graph is defined through the semantics of its "bounds", reusing RDF semantics as is (as we currently do).

In this new semantics, a strict RDF-star graph (i.e. one that contains embedded triples) has no exactly equivalent RDF graph, so it still can not be conveyed exactly using RDF syntaxes ~~(but we do not rely anymore on hidden predicates for that)~~. However, either of the two "bounds" can be used to approximate the RDF-star graph in legacy RDF. The "lower bound" will produce correct but incomplete inferences. The "upper bound" will produce complete inferences, with a few spurious (but generally harmless) ones.

**edited** since the "upper bound" entails the RDF-star graph, the initial version of this proposal still required RDF-star systems to "decode" embedded triples from their RDF descriptions, with all the trouble generated by pathological cases. So I re-introduced one hidden IRI to avoid that. Note however that the hidden IRI is now just a type; the description of the embedded triple is now mostly "visible" as RDF (as an extension of standard reification) -- although it does not entail, nor is entailed by, the embedded triple itself.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/pull/127.html" title="Last updated on Mar 11, 2021, 5:39 PM UTC (75e2b13)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/127/3c42e4d...75e2b13.html" title="Last updated on Mar 11, 2021, 5:39 PM UTC (75e2b13)">Diff</a>


cg-spec/editors_draft.html
```diff
@@ -1779,87 +1779,80 @@ <h2>RDF-star Semantics</h2>
...
+        <li>Unless |G| is an <a>RDF graph</a>, |G1| does not <a>entail</a> |G|, which does not <a>entail</a> |G1| ∪ |G2|.</li>
+        <li>|G1| is the biggest <a>RDF graph</a> <a>entailed</a> by |G|.</li>
+        <li>|G1| ∪ |G2| is the smallest <a>RDF graph</a> <a>entailing</a> |G|.</li>
+      </ul>
 
-        <p>The exchange of a mapped graph |unstar|(|G|) using standard RDF concrete syntaxes, with non-standard <a>IRIs</a> in place of the special properties, is however possible and useful when communicating with legacy RDF systems. For those systems, the special properties have no special meaning, so using non-standard <a>IRIs</a> makes no difference to them.</p>
+      <p>Informally, |G1| and |G1| ∪ |G2| can be thought of as the ``RDF lower and upper bounds'' of the <a>RDF-star graph</a> |G|. They can be used to handle <a>RDF-star graph</a> in legacy RDF systems. More specifically, representing |G| as |G1| will yield correct, but incomplete, reasoning; representing |G| as |G1| ∪ |G2| will yield complete reasoning, but producing spurious inferences (making it incorrect, although these spurious inferences would generally be considered harmless).</p>
```
TallTed, line 1855:
```suggestion
      <p>Informally, |G1| and |G1| ∪ |G2| can be thought of as the "RDF lower and upper bounds" of the <a>RDF-star graph</a> |G|. They can be used to handle <a>RDF-star graphs</a> in legacy RDF systems. More specifically, representing |G| as |G1| will yield correct, but incomplete, reasoning; representing |G| as |G1| ∪ |G2| will yield complete reasoning, but include spurious inferences (making it incorrect, although these spurious inferences would generally be considered harmless).</p>
```

cg-spec/editors_draft.html
```diff
@@ -1841,18 +1842,22 @@ <h2>Entailment of <a>RDF-star graphs</a></h2>
...
+
+      <p>Note however that the IRI `x:EmbeddedTriple` is not specified by this document, which means that there is no <em>standard</em> way to entirely serialize a non-empty |G2|. In other words, there is no standard way to convey an <a>RDF graph</a> <a>entailing</a> a strict <a>RDF-star graph</a> (i.e. one that contains at least one <a>embedded triple</a>). RDF-star aware systems are therefore not required by this specification to inspect <a>RDF triples</a> in order to reconstruct the <a>embedded triples</a> that would have been entailed, which avoids them the trouble of handling pathological corner cases. For example, what should be done if `rdf:subject` is missing? If two `rdf:subject`s are provided? If `rdf:subject` and `rdf:quotedSubject` are not consistent with each other?</p>
```
TallTed, line 1858:
```suggestion
      <p>Note however that the IRI `x:EmbeddedTriple` is not specified by this document, which means that there is no <em>standard</em> way to entirely serialize a non-empty |G2|. In other words, there is no standard way to convey an <a>RDF graph</a> <a>entailing</a> a strict <a>RDF-star graph</a> (i.e., one that contains at least one <a>embedded triple</a>). RDF-star–aware systems are therefore not required by this specification to inspect <a>RDF triples</a> in order to reconstruct the <a>embedded triples</a> that would have been entailed, which saves them the trouble of handling pathological corner cases. For example, what should be done if `rdf:subject` is missing? If more than one `rdf:subject` is provided? If `rdf:subject` and `rdf:quotedSubject` are not consistent with each other?</p>
```

pchampin:
See also the discussion on the mailing list:
https://www.w3.org/mid/828395cf-d9e4-cad3-c051-c5ba88d27366@ercim.eu

Closing this PR, as it does not seem to convince anyone but me :-)

Issue #126, opened 2021-03-05
Clarification section required for the new built-in functions


hartig:
As requested by @lisp [on the mailing list](https://lists.w3.org/Archives/Public/public-rdf-star/2021Mar/0020.html), it would be helpful for the readers of the spec to have a clarification section that illustrates the usage of the new built-in functions (see PR #118), including corner cases related to the creation of triples with blank nodes and the relationship to queries over named graphs within a queried RDF-star dataset.

hartig:
PR #129 adds the clarification section.

PR #129, opened 2021-03-07
Adds a clarification section for the new built-in functions

hartig:
This PR addresses #126

There is a new Section 4.4.6 with examples of the new built-in functions and a discussion of the interplay with the BNODE function and GRAPH clauses.

Notice that my plan is to move the first three examples into the overview/primer part of the report (after creating a proper structure for that part). Hence, after that, what will remain in this new Section 4.4.6 is the discussion of the special cases. However, for the moment I have put it all in this section in order to get feedback on the examples and on the discussion.

/cc @pchampin @afs @gkellogg @TallTed

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/pull/129.html" title="Last updated on Mar 9, 2021, 5:19 PM UTC (13bde09)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/129/3c42e4d...13bde09.html" title="Last updated on Mar 9, 2021, 5:19 PM UTC (13bde09)">Diff</a>


cg-spec/editors_draft.html
```diff
@@ -1420,6 +1420,112 @@ <h2>isTRIPLE</h2>
 
       </section>
 
+      <section class="informative">
+        <h2>Examples and Discussion</h2>
+
+        <p>By the evaluation semantics of SPARQL-star as defined in <a href="#evaluation-semantics"></a>, the five new built-in functions defined above can be used in exactly the same manner as <a data-cite="SPARQL11-QUERY#SparqlOps">the functions defined in the SPARQL 1.1 spec</a><!--, which includes using them within GRAPH clauses that focus on querying named graphs with a queried dataset-->. For instance, the following SPARQL-star query retrieves every <a>asserted triple</a> that, both, is contained in the <a>default graph</a> of the queried <a>RDF-star dataset</a> and has an <a>embedded triple</a> as its <a>subject</a>.</p>
```
TallTed, line 1426:
```suggestion
        <p>By the evaluation semantics of SPARQL-star as defined in <a href="#evaluation-semantics"></a>, the five new built-in functions defined above can be used in exactly the same manner as <a data-cite="SPARQL11-QUERY#SparqlOps">the functions defined in the SPARQL 1.1 spec</a><!--, which includes using them within GRAPH clauses that focus on querying named graphs with a queried dataset-->. For instance, the following SPARQL-star query retrieves every <a>asserted triple</a> that is both contained in the <a>default graph</a> of the queried <a>RDF-star dataset</a> and has an <a>embedded triple</a> as its <a>subject</a>.</p>
```

cg-spec/editors_draft.html
```diff
@@ -1420,6 +1420,112 @@ <h2>isTRIPLE</h2>
...
+          <!--
+          SELECT ?t WHERE {
+            ?s ?p ?o .
+            BIND( TRIPLE(?s,?p,?o) AS ?t )
+            FILTER( isTRIPLE(SUBJECT(?)) )
+          }
+          -->
+        </pre>
+
+        <p>Instead of accessing the subject of the triples in the FILTER after the BIND clause, as done in the previous query, a semantically equivalent variation of this query may apply the FILTER directly on the variable&nbsp;<code>?s</code> as follows.</p>
```
TallTed, line 1441:
```suggestion
        <p>Instead of accessing the subject of the triples in the FILTER after the BIND clause, as was done in the previous query, a semantically equivalent query may apply the FILTER directly on the variable&nbsp;<code>?s</code>, as follows:</p>
```

cg-spec/editors_draft.html
```diff
@@ -1420,6 +1420,112 @@ <h2>isTRIPLE</h2>
...
+          <!--
+          SELECT ?t WHERE {
+            ?s ?p ?o .
+            BIND( TRIPLE(?s,?p,?o) AS ?t )
+            FILTER( isTRIPLE(?s) )
+          }
+          -->
+        </pre>
+
+        <p>While the triples that the previous example queries bind to variable&nbsp;<code>?t</code> occur in the queried data, the <a href="#triple-function">TRIPLE function</a> can also be used to construct triples that may not be in the data. For instance, for every <a>asserted triple</a> in the <a>default graph</a> of the queried <a>RDF-star dataset</a>, the following query uses the subject of this triple to construct another triple and bind it to the variable&nbsp;<code>?t</code>. Notice that the result of this query contains as many <a>SPARQL-star solution mappings</a> as there are asserted triples in the queried graph, and there may be duplicates in the result if multiple asserted triples have the same subject.</p>
```
TallTed, line 1456:
```suggestion
        <p>While the triples that the previous example queries bind to variable&nbsp;<code>?t</code> occur in the queried data, the <a href="#triple-function">TRIPLE function</a> can be used to construct triples that may not be in the data. For instance, the following query uses the subject of every <a>asserted triple</a> in the <a>default graph</a> of the queried <a>RDF-star dataset</a>, to construct another triple and bind it to the variable&nbsp;<code>?t</code>. Notice that the result of this query contains as many <a>SPARQL-star solution mappings</a> as there are asserted triples in the queried graph, and there may be duplicates in the result if multiple asserted triples have the same subject.</p>
```

cg-spec/editors_draft.html
```diff
@@ -1420,6 +1420,112 @@ <h2>isTRIPLE</h2>
...
+          PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+
+          SELECT ?t WHERE {
+            ?s ?p ?o .
+            BIND( TRIPLE(?s,rdf:type,rdfs:Resource) AS ?t )
+          }
+          -->
+        </pre>
+
+        <p>There are cases in which the effects of using the new TRIPLE function are not immediately obvious just from its definition itself. In particular, this concerns the interplay of this function with the <a data-cite="SPARQL11-QUERY#func-bnode">BNODE function</a> when used within <a data-cite="SPARQL11-QUERY#queryDataset">GRAPH clauses</a>. In the following, we discuss such cases.</p>
```
TallTed, line 1473:
```suggestion
        <p>There are cases where the effects of using the new TRIPLE function are not immediately obvious just from its definition, particularly when considering the interplay of this function with the <a data-cite="SPARQL11-QUERY#func-bnode">BNODE function</a> when used within <a data-cite="SPARQL11-QUERY#queryDataset">GRAPH clauses</a>. Some of these cases are discussed below.</p>
```

cg-spec/editors_draft.html
```diff
@@ -1420,6 +1420,112 @@ <h2>isTRIPLE</h2>
...
...use collapses these |n| mappings into a single mapping that is defined for the six variables introduced in the SELECT clause (if |n|=0, the query result is the empty set). Each of these six variables is mapped to an integer-typed literal with the following values. For <code>?t1Count</code>, <code>?t2Count</code>, and <code>?t3Count</code>, the value is&nbsp;|n|, respectively, because the variables <code>?t1</code>, <code>?t2</code>, and <code>?t3</code> are bound in each of the |n| solution mappings produced for the GRAPH clause. For the variable&nbsp;<code>?t1DistCount</code>, the value is |n| as well. The reason for the latter is that the BIND clauses are evaluated |n| times (once for each named graph) and the subexpression BNODE() always creates a different new blank node; thus, the triples created for variable <code>?t1</code> are all different from one another. The same is also the case if the BNODE function is used with an argument; i.e., the value of <code>?t2DistCount</code> is |n| too. In contrast, the variable&nbsp;<code>?t3</code> is mapped to the same triple within each evaluation of the third BIND clause and, thus, the value for <code>?t3DistCount</code> is&nbsp;1.</p>
```
TallTed, line 1498:
```suggestion
        <p>When evaluated over an <a>RDF-star dataset</a> that contains |n| <a>named graphs</a>, where |n|>0, the GRAPH clause results in one <a>SPARQL-star solution mapping</a> per named graph. The SELECT clause collapses these |n| mappings into a single mapping that is defined for the six variables introduced in the SELECT clause. (If |n|=0, the query result is the empty set.) Each of these six variables is mapped to an integer-typed literal with values as follows. For each of <code>?t1Count</code>, <code>?t2Count</code>, and <code>?t3Count</code>, the value is&nbsp;|n|, because the variables <code>?t1</code>, <code>?t2</code>, and <code>?t3</code> are bound in each of the |n| solution mappings produced for the GRAPH clause. For the variable&nbsp;<code>?t1DistCount</code>, the value is also |n|, because the BIND clauses are evaluated |n| times (once for each named graph) and the subexpression BNODE() always creates a different new blank node; thus, the triples created for variable <code>?t1</code> are all different from one another. This is also the case if the BNODE function is used with an argument; i.e., the value of <code>?t2DistCount</code> is |n| too. In contrast, the variable&nbsp;<code>?t3</code> is mapped to the same triple within each evaluation of the third BIND clause; thus, the value for <code>?t3DistCount</code> is&nbsp;1.</p>
```

cg-spec/editors_draft.html
```diff
@@ -1420,6 +1420,112 @@ <h2>isTRIPLE</h2>
...
+
+        <p>Notice that the BNODE function has to be used to create a blank node; simply writing a blank node directly in an expression <a data-cite="SPARQL11-QUERY#rExpression">is not permitted</a>.</p>
+
+        <p>Notice also that the given example query is special in the following two ways: its GRAPH clause does not actually consider the content of the named graphs and the expressions in the BIND clauses do not contain variables. As another example, the GRAPH clause and the BIND clauses in the following query do.</p>
```
TallTed, line 1502:
```suggestion
        <p>Notice as well that the preceding query is special in two ways: its GRAPH clause does not actually consider the content of the named graphs; and the expressions in the BIND clauses do not contain variables. In contrast, the GRAPH clause and the BIND clauses in the following query do.</p>
```

cg-spec/editors_draft.html
```diff
@@ -1420,6 +1420,112 @@ <h2>isTRIPLE</h2>
...
+            GRAPH ?g {
+              ?g rdf:type rdfg:Graph .
+              BIND( TRIPLE(?g, rdf:type, rdfs:Resource) AS ?t1 )
+              BIND( TRIPLE(?x, rdf:type, rdfs:Resource) AS ?t2 )
+            }
+          }
+          -->
+        </pre>
+
+        <p>For every RDF-star dataset with |n|>0 named graphs, the result of this query consists of a single SPARQL-star solution mapping that is defined for the three variables introduced in the SELECT clause. The value of <code>?t1Count</code> is the number of named graphs in the queried dataset that contain a triple of the form (|u|, <code>rdf:type</code>, <code>rdfg:Graph</code>) such that |u| is the name of the named graph in the dataset. The value of <code>?t2Count</code> is always&nbsp;0&nbsp;(zero). The reason for the latter is that evaluating the expression of the second BIND clause results in an error because the variable&nbsp;<code>?x</code> is not bound in the scope of this evaluation. As a side note, the same would also be the case for the first BIND clause if the variable&nbsp;<code>?g</code> was not mentioned inside the GRAPH clause (e.g., if the given triple pattern was not there).</p>
```
TallTed, line 1525:
```suggestion
        <p>For every RDF-star dataset with |n|>0 named graphs, the result of this query consists of a single SPARQL-star solution mapping that is defined for the three variables introduced in the SELECT clause. The value of <code>?t1Count</code> is the number of named graphs in the queried dataset that contain a triple of the form (|u|, <code>rdf:type</code>, <code>rdfg:Graph</code>) such that |u| is the name of the named graph in the dataset. The value of <code>?t2Count</code> is always&nbsp;0&nbsp;(zero), because evaluating the expression of the second BIND clause results in an error, because the variable&nbsp;<code>?x</code> is not bound in the scope of this evaluation. As a side note, evaluating the expression of the first BIND clause would also result in an error, if the variable&nbsp;<code>?g</code> was not mentioned inside the GRAPH clause (e.g., if the given triple pattern was not there).</p>
```

cg-spec/editors_draft.html
```diff
@@ -1420,6 +1420,112 @@ <h2>isTRIPLE</h2>
...
+          <!--
+          SELECT ?t WHERE {
+            ?s ?p ?o .
+            BIND( TRIPLE(?s,?p,?o) AS ?t )
+            FILTER( isTRIPLE(?s) )
+          }
+          -->
+        </pre>
+
+        <p>While the triples that the previous example queries bind to variable&nbsp;<code>?t</code> occur in the queried data, the <a href="#triple-function">TRIPLE function</a> can be used to construct triples that may not be in the data. For instance, the following query uses the subject of every <a>asserted triple</a> in the <a>default graph</a> of the queried <a>RDF-star dataset</a>, to construct another triple and bind it to the variable&nbsp;<code>?t</code>. Notice that the result of this query contains as many <a>SPARQL-star solution mappings</a> as there are asserted triples in the queried graph, and there may be duplicates in the result if multiple asserted triples have the same subject.</p>
```
TallTed, line 1456:
```suggestion
        <p>While the triples that the previous example queries bind to variable&nbsp;<code>?t</code> occur in the queried data, the <a href="#triple-function">TRIPLE function</a> can be used to construct triples that may not be in the data. For instance, the following query uses the subject of each <a>asserted triple</a> in the <a>default graph</a> of the queried <a>RDF-star dataset</a> to construct another triple, and bind it to the variable&nbsp;<code>?t</code>. Notice that the result of this query contains as many <a>SPARQL-star solution mappings</a> as there are asserted triples in the queried graph, and there may be duplicates in the result if multiple asserted triples have the same subject.</p>
```

cg-spec/editors_draft.html
```diff
@@ -1420,6 +1420,112 @@ <h2>isTRIPLE</h2>
...
+          PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+
+          SELECT ?t WHERE {
+            ?s ?p ?o .
+            BIND( TRIPLE(?s,rdf:type,rdfs:Resource) AS ?t )
+          }
+          -->
+        </pre>
+
+        <p>There are cases where the effects of using the new TRIPLE function are not immediately obvious just from its definition, particularly when considering the interplay of this function with the <a data-cite="SPARQL11-QUERY#func-bnode">BNODE function</a> when used within <a data-cite="SPARQL11-QUERY#queryDataset">GRAPH clauses</a>. Some of these cases are discussed below.</p>
```
jeenbroekstra, line 1473: For what it's worth I find these cases contrived. It's basically an explanation that says: "if you write this weird artificial construct, here's what happens". All of the elements mentioned are well-understood parts of the language, how they interact is not in any way ambiguous, it's complex only because it is doing some non-trivial combination of language features. 

I have no objection to including it in the editor draft, and the explanation as it stands is a clear and concise one, but I'd be a little wary of adding as part of an introductory / primer text.

hartig: @jeenbroekstra I agree. Note that there was a request to have a discussion of such contrived cases in this part of the spec, and my intention is to keep it here. The only thing that I would move to the overview/primer part of the report are the three examples that come before the discussion of these contrived cases (i.e., Examples 12-14).

lisp:
i have, but i need to think about it. i was expecting to have time until friday.
is that too late?

lisp:
ex 12: as expressed, the presence is in the default dataset, not necessarily the default graph.

ex 13: is the bind not gratuitous? this symmetry may simplify the subsequent text, but will it not also confuse?

ex 14: also has the default dataset as target.
 “uses just the subject of each asserted…”
 perhaps call out that count and count distinct wmay yield different values

ex15: “The SELECT clause aggregates…”
the discussion should proceed further.
for instance, it could introduce the case where each graph contained the single statement
```
 :s :p _:blank
```
in an a-typical case, in which the graphs contained statements which include the _same_ blank node and the bind clause is
```
 BIND ( TRIPLE(?s, ?p, ?o) as ?t3)
```
i suspect that the aggregate count would also be 1, as ?o is bound to the same blank node.
i hope i am not surprised.

given which, it could be informative to discuss the consequences of an process which operated on two initially empty datasets with an initial update operation on the order of
```
graph ?g {
 ?s ?p ?node .
 ?s ?p _:node }
}
where {
 values ?g {<http://ex.org/1> <http://ex.org/2>}
 bind (bnode(’node') as ?node)
 bind (<http://ex.org/p> as ?p)
 bind (TRIPLE(?node, ?p, ‘o’) as ?s)
}
```
exported that dataset, imported the intermediate document into the second dataset and then executed queries on each dataset on the order of 
```
select  (count (?node ) as ?count) (count (distinct ?o ) as ?distinctCount)
where {
graph ?g {
 <<?node <http://ex.org/p> ‘o’>> <http://ex.org/p> ?node .
}
````
or

```
select (count (?node ) as ?count) (count (distinct ?o ) as ?distinctCount)
where {
graph ?g {
 ?s <http://ex.org/p> ?node .
}
```
at issue of the scope and extent of the bindings between blank node labels and the respective nodes as well as the extent 
the consequences for a node which is not in any graph for the results of those queries is not yet self-evident.
for example, would the count in the second query be 1 or 0?
there could be some value to explaining this.

if, as has been suggested, the answers follow from the semantics rather than from the function which this sections describes, the discussion could well be elsewhere in the document, but in that case, a reference from this section would serve the readers.

**Comment by afs** [comment]
@lips What is a "default dataset"? (I think the example 12 text is correct)

**Comment by TallTed** [comment]
@lisp
> the graphs contained statements which include the _same_ blank node

How is this scenario possible?  My understanding of blank (i.e., unnamed, unidentified) nodes is that their pseudo-identification is etheric beyond the bounds of an enclosing graph or result set.  They may get assigned temporary, etheric, stick-on labels bearing identifiers which persist for the duration of a query and its result set, but those identifiers cannot be used outside of that query and result set.

Yes, you might describe a node without an identifier (URI) of its own with the same predicates/attributes and objects/values in multiple graphs — but how can you or any SPARQL querent determine that the *same* node is being described in those graphs?

Open World says that anything unstated is unknown.  This means that an attribute/predicate that is left out could have any value/object — and, if it had been included, it could be different in every one of the graphs in question, and thus the unidentified node be different.

For instance, you could have 150 graphs, each containing a description of a yellow 2015 Chevrolet Camaro, where the objects/values of all predicates/attributes therein are identical — but none of those graphs include the VIN which, by definition, is different for each Camaro that rolls off the assembly line.  Unfortunately for your scenario, every description would have had a different VIN, because each of the 150 graphs actually describes a different Camaro — even though they *appear* to be the same based on the attributes/predicates in those graphs.

So, in your scenario, "the aggregate count" *cannot* "be 1" — it _must_ be the number of graphs which contain that blank node, because that is the number of etheric, temporary blank-node "identifiers" (which as stated above aren't identifiers in the larger sense; they are not URIs, and they are only useful within the context of this result set) the SPARQL engine must assign for purposes of that SPARQL result set.

**Comment by lisp** [comment]
> What is a "default dataset"? (I think the example 12 text is correct)

13.2, the section in the sparql 1.1 recommendation  about "specifying rdf datasets", describes this as "any dataset that the query service would use if no dataset description is provided in a query."

**Comment by lisp** [comment]
> How is this scenario possible? My understanding of blank (i.e., unnamed, unidentified) nodes is that their pseudo-identification is etheric beyond the bounds of an enclosing graph or result set. They may get assigned temporary, etheric, stick-on labels bearing identifiers which persist for the duration of a query and its result set, but those identifiers cannot be used outside of that query and result set.

the rules which govern the scope and extent of the binding of a label to a blank node are not the same as the ones which govern the extent of the node.

**Comment by afs** [comment]
Thank you for the clarification.

The whole query is executed against a dataset and indeed here, there is no FROM,FROM NAMED.

The text "contained in the default graph" is correct for the example query because the "?s ?p ?o" is only against the default graph (of the dataset being queried).

**Comment by lisp** [comment]
... unless, as we have been made aware by some users, the default graph of the default dataset happens to merge the contents of the named graphs.

**Comment by afs** [comment]
How the default graph comes about is not of concern to SPARQL query execution. 

The query engine is responsible for the abstraction of the default graph as a set of triples. Sets have a "contains" relationship and "?s ?p ?o" matches all the triples in that set.

**Comment by lisp** [comment]
as it can result from merging the named graphs, its origin can be of concern.

**Comment by afs** [comment]
It may be of concern but then this isn't the right query to ask! Use `GRAPH ?g { ... }` to the containing named graph(s).

For example 12, the text as explanation about the query matching the default graph is correct.

**Comment by lisp** [comment]
the concern raised in this thread is as to the behaviour of blank nodes constituent to embedded triples which are terms in statements in graphs.
the proposed correction seeks 
- either to exclude any possible merged graphs from consideration at this point, or
- to account for their effect, and
- to be consistent in phrasing with later examples which need to incorporate graphs.

while the current phrasing in not incorrect, it may also be incomplete, for some cases which resolves the uncertainties raised.

**Comment by afs** [comment]
See https://www.w3.org/TR/sparql11-query/#sparqlDataset - there is one set of blank nodes for the dataset (the data) with no regard to their use in any graph.

> either to exclude any possible merged graphs from consideration at this point,

SPARQL (1.0) was intentionally spec'ed to include this possibility.

It does not matter. There is a default graph. It is an RDF graph. Its origin is outside the SPARQL query specification.

> to account for their effect

I don't know what this means that is not covered by the definition of an RDF dataset as referenced.

Please provide an example that can not be determined by the current text with alternative possibilities of interpretation.

**Comment by lisp** [comment]
the cited passage from the sparql recommendation references a definition for "merge" : https://www.w3.org/TR/rdf-mt/#defmerge .
that definition describes a result which is constructed such that it comprises "equivalent graphs that share no blank nodes".
it is not clear, how that resolves the concern about any "embedded blank nodes" which may be in named graphs which are merged by an implementation which merges all named graphs in the default graph in a default dataset.

**Comment by afs** [comment]
I am still not clear what the concern with regard to RDF-star is. 

Please provide a concrete example (data+query) and say what step in the evaluation of the example in the preview is unclear.

"merge" is well defined for both individual graphs and datasets.
(The discussion is in section 13.1).

You will find that several implementations that provide the "merge" with "union" because they use the implementation knowledge that blank nodes across the dataset are distinct IFF they have distinct internal identifiers (not blank node labels in syntax). Jena uses UUID strings for blank nodes uniquely generated by the RDF syntax parsers.

Dydra does not support the default graph being the composite of named graphs does it?

TriG:
```
PREFIX : <http://example/>
GRAPH :g1 { _:a :p :z1 }
GRAPH :g2 { _:a :p :z2 }
```
and we know there is exactly one blank node because TriG scopes labels to the document.

But we can not deduce that from these two files: 
```
PREFIX : <http://example/>
GRAPH :g1 { _:a :p :z1 }
```
```
PREFIX : <http://example/>
GRAPH :g2 { _:a :p :z2 }
```
where a merge will have two blank nodes.

A different implementation where two graphs have been separate read in, and blank nodes are, for example, internally distinguished with system identifiers per graph will rename the blank nodes apart on merge (e.g. prepend the graph name to the internal identifier). This is an implementation matter and not related to SPARQL because SPARQ query pattern matching does not start until there is a valid dataset however that comes about.

The RDF1.1 reference is:
https://www.w3.org/TR/rdf11-mt/#shared-blank-nodes-unions-and-merges

lisp:
> I am still not clear what the concern with regard to RDF-star is. ... "merge" is well defined for both individual graphs and datasets.

> The RDF1.1 reference is:
https://www.w3.org/TR/rdf11-mt/#shared-blank-nodes-unions-and-merges

which passage reiterates the one cited earlier. they concern statements in graphs.

how do the various ways to treat blank nodes when performing the merge or union of graphs when constructing a target dataset apply to the blank nodes which are comprised by embedded triples which are present in statements in named graphs?

lisp:
> Dydra does not support the default graph being the composite of named graphs does it?

in our case, for a query which does not specify a dataset, the default dataset does not combine the named graphs into the default graph.
a query which intends that combination must designate it as the source selector in the default graph clause of the dataset specification

afs:
[RDF 1.1 MT section 4.1](https://www.w3.org/TR/rdf11-mt/#shared-blank-nodes-unions-and-merges) has text not in RDF 1.0 about blank nodes that are known to be shared between graphs in a dataset.

lisp:
> RDF 1.1 MT section 4.1 has text not in RDF 1.0 about blank nodes that are known to be shared between graphs in a dataset.

those descriptions from the rdf semantics document have been cited directly and indirectly at several places in this thread. they describe the extent of blank nodes which are in graphs and the scope and extent of their designators.  
the uncertainty which gave issue to this thread is not related to the behaviour of that category of blank node.

take the case which is described in this rdf-star document as [example 1](https://pr-preview.s3.amazonaws.com/w3c/rdf-star/pull/118.html#example-1).
replace the term `:employee38` with `_:employee38`.
the example suggests that the new triple

    << _:employee38 :jobTitle "Assistant Designer" >>

would also not be asserted. one reading of that characterization is that this new triple is not in any graph.
a similar conclusion could apply to the triples which are constructed by the built-in `TRIPLE` operator.
one consequence of that could be that the blank node in the result triple, which was designated in that expression by `_:employee38`, is likewise not in any graph. if that implication is correct, then it is not self-evident how to read the descriptions from  [RDF 1.1 MT section 4.1](https://www.w3.org/TR/rdf11-mt/#shared-blank-nodes-unions-and-merges) and co. so that they apply to it.

that is, the concern here is, given the reading, that the blank nodes which this operator incorporates into triples are of (or, may be transformed into) a category different from those described by "merge" and co, the text should describe and provide examples which clarify their extent and the scope and extent of their label bindings.
it follows also from that reading, that the examples should include cases which construct, target and merge graphs.
from that follows, in order that the examples exhibit a uniform structure, they could well describe the constitution of the target dataset on all cases, including those where it is the default dataset.

pchampin:
@lisp the confusion comes from the ambiguity of the expression "to be in a graph". In your latest comment, you wrote  "one reading is that (...) this new triple is not in any graph. (...)  `_:employee38` is likewise not in any graph. If that implication is correct ...". I consider that this implication is incorrect, precisely because "in any graph" means very different things in the two parts of this implication. Let me explain. 

In standard RDF, graphs are *sets* of triples. A triple *t* is "in a graph" *g* iff *t* ∈ *g*. An RDF term *u* can not be an element of *g*, so when we say that *u* is "in the graph" *g*, we obviously mean something else, namely that *u* is used in a triple that is an element of *g*.

In RDF-star, triples are sometimes elements of a graph (asserted triples), and sometimes terms of that graph (embedded triples). So saying that a triple is (not) in a graph is ambiguous. We should always specify that "*t* **is asserted in** *g* (formally: *t* ∈ *g*) or "*t* **is embedded in** *g*" (formally: *t* ∈ [constituents][1](*g*)).

In your example above, the triple ``_:employee38 :jobTitle "Assistant Designer"`` is indeed not asserted in the graph, but it is obviously embedded in the graph. In other words, it is in the [constituents][1] of the graph, and likewise, `_:employee38` is also in the constituents of that graph (because constituents is defined recursively).

Granted, whenever the RDF spec says something about the "terms in a graph", one may wonder how this translates to RDF-star: does this concern only the terms appearing as subject, predicate and object of some asserted triple, or does it concern all the constituent terms? I argue that the latter interpretation should be preferred, because the former creates an unspecified gray area of terms that are neither "in the graph", nor totally "out of it" (at least, not as much "out of it" as terms that are not even in the constituent terms).

[1]: https://w3c.github.io/rdf-star/cg-spec/2021-02-18.html#dfn-constituent

lisp:
> whenever the RDF spec says something about the "terms in a graph", one may wonder how this translates to RDF-star: does this concern only the terms appearing as subject, predicate and object of some asserted triple, or does it concern all the constituent terms?

this _is_ the question which i have been posing for more than a week.
and, while my uncertainty as to how to manage blank nodes which are in a graph while their containing triple is not, may lead me to pose the question, that is a separate question.
whatever its answer may be, the concern remains, that the description of an operator defined to yield such a result should either include the answer or identify its location.

afs:
There could be clarification that RDF merge, and the blank node naming apart, applies to embedded triples. 

Personally, the language in MT 1.1, while not designed for RDF-star says that. Renaming keeps the shape of the graph (isomorphism) so renaming happens for renaming terms (blank nodes) in embedded triples.

pchampin:
@lisp now that we have [an explicit text about merge](https://w3c.github.io/rdf-star/cg-spec/editors_draft.html#merging) in the spec, do you consider the content of this PR to be sufficient?

lisp:
that sentence is sufficiently definitive and that it be there was the goal of the objection.

pchampin:
this was discussed during today's call https://w3c.github.io/rdf-star/Minutes/2021-04-09.html#x069

@hartig I propose that we merge this PR as is for the next public draft.
Moving the first examples into the Overview can be done afterwards (see also #155).

hartig:
@pchampin sounds good. Thanks!

gkellogg (approved):

PR #130, opened 2021-03-12
Make manifest bases not include the file extension

gkellogg:
(e.g., `<https://w3c.github.io/rdf-star/tests/sparql/eval/manifest#sparql-star-update-3>`).

Required a bit of post-serialization hackery of the manifest.jsonld contexts.

Also, includes building the JSON-LD manifests in the Makefile target.

cc/ @afs


tests/index.html
```diff
@@ -1 +1,103 @@
...
+        <img alt='W3C' height='48' src='http://www.w3.org/Icons/w3c_home' width='72'>
+      </a>
+    </p>
+    <h1>RDF-star Test Suite</h1>
+    <p>This manifest loads additional manifests for specific behavior tests for
+      <a href="https://w3c.github.io/rdf-star">RDF-star</a></p>
+
+    <p>This is an HTML version of a test manifest.
+      Alternate versions of the manifest may be found at
+      <a href="manifest.jsonld">manifest.ttl</a>, and
```
TallTed, line 22:
```suggestion
      <a href="manifest.ttl">manifest.ttl</a>, and
```

tests/index.html
```diff
@@ -1 +1,103 @@
...
+      <a href="manifest.jsonld">manifest.jsonld</a>.</p>
+
+    <p>The RDF-star Test Suite is a set of tests that can
+    be used to verify Processor conformance.</p>
+
+    <h2>General instructions for running the RDF-star Test suites</h2>
+
+    <p>Tests are run broadly the same as those for relavant
```
TallTed, line 30:
```suggestion
    <p>Tests are run broadly the same as those for relevant
```

tests/index.html
```diff
@@ -1 +1,103 @@
...
+      <li>See <a href="semantics/README">general guidance for running Semantics tests</a>.</li>
+    </ul>
+
+    <p>See the <a href="../reports/">implementation</a> report for guidance on reporting test
+      results using the EARL format.</p>
+
+    <h1>Contributing Tests</h1>
+
+    <p>If you would like to contribute a new test or a fix to an existing test,
+      please follow these steps:</p>
```
TallTed, line 65: Is there a reason not to suggest using github's largely automatic web interface, as opposed to what follows which appears to mandate using the manual clone/edit/PR process?

gkellogg: If there's a simpler way, please recommend. I'm not familiar with a simple way to make all the changes typically needed to do such a change. There is a way to add or edit individual files, perhaps these could be coordinated into a single commit/PR. If there's a description of a purely web-based workflow, please let me know, or suggest an edit.

The intention of this is to lay out the steps typically necessary to submit such a change, not to mandate that as being the only way to make such a change.

TallTed: Yes, it's possible to change multiple files before submitting a PR, all through the web interface (which is how I operate).

Editing the first file automatically (and all but invisibly) triggers creation of a branch/clone. The web interface then suggests creating a PR immediately, but this is not required. Without closing that tab, I control-click on the automatically created branch name, and open it in a new tab (so I can most easily later use the auto-PR tools in the original tab). I again control-click on the other document(s) I want to change for this PR, opening each in a new tab. I make all the other changes connected to the first, and write them directly to the auto-created branch, closing each tab as the edits on that file are done. When all changes are done, I reload the tab that first prompted me to create a PR -- and that PR will now include the full DIFF between this branch and the original.

I can do this again, to create branch-2, etc., to build another PR with multiple files being changed, based on the original branch -- so, without the changes in the PR above.

The above isn't written with each and every step, because it's from memory, not from going through the motions, so I wouldn't include it in the README.  But I'd change "please follow these steps" to something along the lines of, "Use your preferred method to create and submit a Pull Request, if you know what would fix the problem you see, or open an Issue, if you see a problem, but don't know how to fix it."

gkellogg:
```suggestion
please use either the [GitHub web interface](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request), or use the following steps using a git client, or equivalent:</p>
```

gkellogg: I think that change adequately describes both your workflow and the remote client mechanism. Feel free to suggest an improvement.

pchampin (approved):
I added a simplification of make-html-manifests.py, to now rely on make-jsonld-manifests.rb for JSON-LD generation (separation of concerns). This also fixes some broken links in the entry list of the HTML manifests.

Would it make sense to run `make` in the `tests` directory in CI?

gkellogg:
We could run make just for CI purposes, or have a different action to run make an commit any updated files, to ensure that they get built on any change. This might be expensive if extended to reports without improved dependency checks. 

Certainly, for now, running it as part of CI might be useful. It might even be considered an error if any files are out of date.

tests/index.html
TallTed: Minor tweaks to grammar and a bit of wording 
```suggestion
    <p>If you would like to contribute a new test or a fix to an existing test,
please submit a Pull Request using either the <a 
href="https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request">GitHub web interface</a>, 
the steps below with a git client, or your preferred equivalent.</p>

File: tests/turtle/syntax/manifest.ttl [hunk 4]
```diff
@@ -3,6 +3,7 @@
 ## [1] https://www.w3.org/Consortium/Legal/2008/04-testsuite-license
 ## [2] https://www.w3.org/Consortium/Legal/2008/03-bsd-license
 
+BASE           <manifest>
```
afs, line 6: Don't we want to fix the absolute URI?

`BASE <https://w3c.github.io/rdf-star/tests/turtle/syntax/manifest>`

gkellogg: This was the least change, as the manifests were previously treated as location relative. And, if we choose to eventually move the manifests to a different location. But I can update to use the absolute IRI, if you like.

afs:
My tests run _extremely_ slowly now :-)

If `BASE` changes, then the file locations change to be the web. I think we want them local.

If `PREFIX :` changes, only the URL of tests (that goes into `earl:test`) changes and the suite runs off local files.

That leaves a separate issue about the URL of the manifests file itself. It is `<>` currently, while tests are `:some-test`.
This is what the "rdf-tests" are like. The fact the manifest starts `<> a mf:Manifest` does not get in the  way because only the test URL matters.

```
PREFIX :       <https://w3c.github.io/rdf-star/tests/sparql/syntax/manifest#>
```
I'm happy to do the PR .


There are other more elaborate ways (not a proposal!):
```
PREFIX local:   <.>
BASE <https://w3c.github.io/rdf-star/tests/sparql/syntax/manifest>
PREFIX : <#>
```
and all files are `local:sparql.rq` but this isn't supposed to be a resolver test suite!

gkellogg:
My own runner intercepts remote load requests that match the repository location and attempts to satisfy them locally.

But, one advantage of the previous `BASE <manifest>` would be to maintain the directory relative nature of file loading. That is the pattern we used for jsonld tests.

afs:
There are two aspects - file location and test URIs. Having absolute BASE and relative prefix fixes both.

I don't think we want the test URIs if the directory structure changes because the test URIs are in the EARL reports.

`BASE <manifest>` is fine if that helps the HTML generation.

The suggestion is to have fixed URI for the tests entries (not the files for the tests).
```
PREFIX :       <https://w3c.github.io/rdf-star/tests/sparql/syntax/manifest#>
```
is not dependent on base.
```
[ rdf:type         earl:Assertion ;
  earl:assertedBy  <http://jena.apache.org/#arq> ;
  earl:mode        earl:automatic ;
  earl:result      [ rdf:type      earl:TestResult ;
                     dc:date       "2021-03-18+00:00"^^xsd:date ;
                     earl:outcome  earl:passed
                   ] ;
  earl:subject     <http://jena.apache.org/#arq> ;
  earl:test        <https://w3c.github.io/rdf-star/tests/sparql/eval/manifest#sparql-star-pattern-7>
] .
```

gkellogg:
The `PREFIX :` solution works in Turtle, but not so well in JSON-LD. Basically, `@id` is evaluated relative to the base location (`@base` in the context, or the actual file location). And, JSON-LD does not support (for historical reasons) and empty prefix. Generally, you'd like to see something like `"@id": "#ntriples-star-4"` to continue to be relative to `<https://w3c.github.io/rdf-star/tests/n3/syntax/manifest>` along with ` "action": "ntriples-star-syntax-4.nt"` to locate the test file. We _could_ create something like a _local_ prefix to get to local files such as ` "action": "local:ntriples-star-syntax-4.nt"`, but it looks odd.

Using a relative base, works for both JSON-LD and Turtle, and is pretty equivalent to not having an explicit `BASE / @base` specified from a processing point. It's unfortunate that file extension becomes part of the base calculation, but RFC3986 doesn't know about this, and content negotiation to remove format considerations from resources hasn't proved to be practical. IMO, overriding the tail portion of the base calculation is the best solution available that preserves the formatting people typically expect.

That said, `BASE <manifest>` doesn't really affect HTML generation, AFAIK.

Typically, a README (or the HTML version of the manifest) would instruct implementors to run the tests as if they were located relative to `<https://w3c.github.io/rdf-star/tests/>` so that we don't need to use absolute URLs in the manifests, themselves. This can simplify running from a local copy, as long as provisions are taken to use the actual location when reporting. SHACL even goes so far as to use a URN as the manifest base.

afs:
I'm happy to accomodate JSON-LD.

Can we have a non-empty prefix? Does that fix the issue for jSON-LD:
```
PREFIX test: <https://w3c.github.io/rdf-star/tests/n3/syntax#>
...
test:sparql-star-basic-2
...
```
(no `manifest`, no `BASE` needed - seems to work). This makes the location of files relative and the suite can be moved around as you want.

gkellogg:
Why don’t you do a PR with the Turtle bits and I’ll add a comimit to update jsonld.

afs:
OK

PR #131, opened 2021-03-13
Implementation reports

gkellogg:
Just Ruby implementations for right now.

### DIFF
--- a/reports/README.md
+++ b/reports/README.md
@@ -1,0 +1,15 @@
+This is a collection of individual
+[EARL reports](https://www.w3.org/TR/EARL10-Schema/) for
+test subjects claiming JSON-LD processor conformance.
+
+The consolidated report is saved to `index.html` generated
+using the
+[earl-report Ruby gem](https://rubygems.org/gems/earl-report).
+Run it as follows:
+
+```sh
+$ gem install earl-report rdf-turtle json-ld rdf-ordered-repo
+$ rm -f manifests.nt && (cd ..; rake reports/manifests.nt)
+$ earl-report --format json -o earl.jsonld *.ttl
+$ earl-report --json --format html --template template.haml -o index.html earl.jsonld
+```
--- a/cg-spec/editors_draft.html
+++ b/cg-spec/editors_draft.html
@@ -91,6 +91,8 @@
       wgPublicList: "public-rdf-star",
       shortName: "rdf-star",
       group: "rdf-dev",
+      testSuiteURI: "https://w3c.github.io/rdf-star/tests/",
+      implementationReportURI:"https://w3c.github.io/rdf-star/reports/",
       xref: ["RDF11-CONCEPTS", "SPARQL11-QUERY", "RDF11-MT"],
       maxTocLevel: 3,
     };
--- a/reports/.earl
+++ b/reports/.earl
@@ -1,0 +1,5 @@
+---
+:format: :json
+:manifest: manifests.nt
+:bibRef: ! '[[rdf-star]]'
+:name: RDF-star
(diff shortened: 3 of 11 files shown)


reports/README.md
```diff
@@ -0,0 +1,15 @@
+This is a collection of individual
+[EARL reports](https://www.w3.org/TR/EARL10-Schema/) for
+test subjects claiming JSON-LD processor conformance.
+
+The consolidated report is saved to `index.html` generated
+using the
+[earl-report Ruby gem](https://rubygems.org/gems/earl-report).
+Run it as follows:
```
TallTed, line 8:
```suggestion
The consolidated report is generated using the
[earl-report Ruby gem](https://rubygems.org/gems/earl-report),
which saves it to `index.html`.

Run it as follows:
```

pchampin (approved):

afs:
Would it be possible to split this into one for the EARL report generator machinery and one for each report?

I am assuming the overall workflow long-term is that implementers, or a proxy helping them, adds an implementation as a PR that is just about one system. We then have a record of who-what-when. It would be nice if the first reports show how it is done.

afs:
What's `manifests.nt`? I don't think it should be in the checked-in files?

If it is needed, can it be a Turtle file with prefixes so it is a little more viewable?

afs:
Great system!

I tried out the instructions in the README but the cd is breaking things:

```
 >>  rm -f manifests.nt && (cd ..; rake reports/manifests.nt)
rake aborted!
No Rakefile found (looking for: rakefile, Rakefile, rakefile.rb, Rakefile.rb)
```
I guessed and parsed all the TTL files into one NT file.
Then:
```
earl-report: command not found
```
Need to note that .../gems/bin is on the path.

There is an implicit assumption about use of `<https://w3c.github.io/rdf-star/>` -- see comment about making manifest absolute and not depend on an external URI. https://github.com/w3c/rdf-star/pull/130#discussion_r595936487

gkellogg:
> Would it be possible to split this into one for the EARL report generator machinery and one for each report?
> 
> I am assuming the overall workflow long-term is that implementers, or a proxy helping them, adds an implementation as a PR that is just about one system. We then have a record of who-what-when. It would be nice if the first reports show how it is done.

Sure, I can do that. Note that I usually need to tweak the reports to deal with minor, but typical, issues. I may need to be added as a committer to the repo to be able to update such PRs.

> What's `manifests.nt`? I don't think it should be in the checked-in files?
> 
> If it is needed, can it be a Turtle file with prefixes so it is a little more viewable?

I can actually update the .earl file to reference the manifests directly, particularly if BASE is absolute, so we can eliminate this file. I've used it, for example in JSON-LD, to simplify the workflow.


> ```
> earl-report: command not found
> ```
> 
> 
> Need to note that .../gems/bin is on the path.
> 
> 
> There is an implicit assumption about use of `<https://w3c.github.io/rdf-star/>` -- see comment about making manifest absolute and not depend on an external URI. [#130 (comment)](https://github.com/w3c/rdf-star/pull/130#discussion_r595936487)

Yes, and with the addition of a Gemfile in PR #130, I'll update the instructions to `bundle install; bundle exec rake`

pchampin:
> Sure, I can do that. Note that I usually need to tweak the reports to deal with minor, but typical, issues. I may need to be added as a committer to the repo to be able to update such PRs.

Sure. Done.

gkellogg:
Other todo:

* [ ] Add to CI and fail if an issue is detected in one of the constituent earl files.
* [ ] Replace this PR with one for infrastructure, and separate for individual submissions.

pchampin:
This was discussed during today's call:
https://w3c.github.io/rdf-star/Minutes/2021-03-19.html#t03

gkellogg:
Replaced by #137.

PR #133, opened 2021-03-19
manifest.ttl - no BASE; PREFIX trs

afs:
Follow on to #130.

1. Use `PREFIX trs:` ("test rdf star") for the test entries in manifest.ttl
2. Remove BASE usage so local files are accessed

Typical entry: URI is `<area#>`:
```
PREFIX trs:    <https://w3c.github.io/rdf-star/tests/semantics#>
```

### DIFF
--- a/tests/make-jsonld-manifests.rb
+++ b/tests/make-jsonld-manifests.rb
@@ -2,6 +2,9 @@
 require "bundler/setup"
 require 'json/ld'
 require 'rdf/turtle'
+require 'rdf/isomorphic'
+require 'rdf/ordered_repo'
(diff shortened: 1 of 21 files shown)


gkellogg:
@afs I added some missing entries to the N-Triples manifest for bad-4 tests. They weren't showing up in the JSON-LD version due to framing. I presume that these tests were intended to be included.

Also, fixed a bad entry in semantics tests.

The JSON-LD conversion script now validates that the manifests are isomorphic with the Turtle versions.

gkellogg (approved):

PR #140, opened 2021-03-23
adding rdft:TestTurtleEval tests

josd:


gkellogg:
@josd you should join the CG to get past the IPR check. I can manually allow this one, though.

josd:
Thx and now joined https://www.w3.org/community/rdf-dev/join

Issue #135, opened 2021-03-19
write some text WRT RDF-star merge


pchampin:
as decided during call: https://w3c.github.io/rdf-star/Minutes/2021-03-19.html#t04
and in response to https://github.com/w3c/rdf-star/pull/129#issuecomment-799621427

gkellogg:
See #143.

PR #143, opened 2021-03-25
section on merge (address #135)

pchampin:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/pull/143.html" title="Last updated on Apr 7, 2021, 10:07 AM UTC (97107f8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/143/318260c...97107f8.html" title="Last updated on Apr 7, 2021, 10:07 AM UTC (97107f8)">Diff</a>

### DIFF
--- a/cg-spec/editors_draft.html
+++ b/cg-spec/editors_draft.html
@@ -2054,7 +2054,7 @@ SELECT ?claimer WHERE {
     <section id="merging">
       <h2>Merging RDF-star graphs</h2>
 
-      <p>[[[RDF11-MT]]] [[RDF11-MT]] defines the <a>merging</a> of two or more <a>RDF graphs</a> as "[taking their] union after forcing any shared blank nodes, which occur in more than one graph, to be distinct in each graph." It is worth clarifying that, in the case of <a>RDF-star graphs</a>, any <a>blank node</a> in the <a>constituent terms</a> of that graph is governed by the definition above, <em>not</em> only those in the subject or object position of some <a>asserted triple</a>.</p>
+      <p>[[[RDF11-MT]]] [[RDF11-MT]] defines the <a>merging</a> of two or more <a>RDF graphs</a> as "[taking their] union after forcing any shared blank nodes, which occur in more than one graph, to be distinct in each graph." Note that, in the case of <a>RDF-star graphs</a>, any <a>blank node</a> in the <a>constituent terms</a> of that graph is governed by the definition above, <em>not</em> only those in the subject or object position of some <a>asserted triple</a>.</p>
 
       <pre id="merging-example"
         data-transform="updateExample"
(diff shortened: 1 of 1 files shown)


cg-spec/editors_draft.html
```diff
@@ -1867,6 +1867,39 @@ <h2>Entailment of <a>RDF-star graphs</a></h2>
       </ul>
     </section>
 
+    <section id="merging">
+      <h2>Merging RDF-star graphs</h2>
+
+      <p>[[[RDF11-MT]]] [[RDF11-MT]] defines the <a>merging</a> of two or more <a>RDF graphs</a> as "[taking their] union after forcing any shared blank nodes, which occur in more than one graph, to be distinct in each graph." It is worth precising that, in the case of <a>RDF-star graphs</a>, any <a>blank node</a> in the <a>constituent terms</a> of that graph is concerned by the definition above, <em>not</em> only those in the subject or object position of some <a>asserted triple</a>.</p>
```
TallTed, line 1873: Is the almost-duplicate `[[[RDF11-MT]]] [[RDF11-MT]]` necessary? It looks a bit odd in the rendering, and more so here.
```suggestion
      <p>[[[RDF11-MT]]] [[RDF11-MT]] defines the <a>merging</a> of two or more <a>RDF graphs</a> as "[taking their] union after forcing any shared blank nodes, which occur in more than one graph, to be distinct in each graph." It is worth noting that, in the case of <a>RDF-star graphs</a>, any <a>blank node</a> in the <a>constituent terms</a> of that graph is concerned by the definition above, <em>not</em> only those in the subject or object position of some <a>asserted triple</a>.</p>
```

gkellogg: [[[RDF11-MT]]] should be expanded by ReSpec to (a reference to) the title of the specification, while [[RDF11-MT]] is a reference to the bibliographic entry.

TallTed: Ah, OK. It looked odder in the DIFF than the PREVIEW, probably because of so many up-arrows for the insertions; I should have looked at both.

I think the one word change I suggested is still necessary. (If I've misunderstood what was meant by the non-word "precising", I'm happy to have another word substitute for my "noting".)

pchampin: Oh.. sorry about that. What I meant was "clarifying", which I think is better than "noting" in this context. Would that work for you?

TallTed: "clarifying" is fine.

afs (approved):

cg-spec/editors_draft.html
lisp: where you say "is concerned", consider using instead "is governed by".

TallTed (approved):

PR #144, opened 2021-03-25
Ruby instability

gkellogg:
For some reason, we've been getting segmentation faults on some runs relating to running reports. This PR drops the Ruby version from 3.0 to 2.7, uses an updated version of the earl-report gem which pulls in fewer supporting files, and does not clean the reports before building, so it should only be triggered when a new earl submission is received that hasn't been used to run reports.


gkellogg:
Note that the earl.ttl still uses the undefined `earl:Report` and `earl:assertions` terms, although in JSON-LD, `assertions` is a reverse property on `earl:test`. It's still useful for organizing the Turtle output, even though it is apocryphal.

Formerly, the top-level report also used `earl:assertions`, not they apocryphal `mf:report`.

pchampin (approved):

reports/index.html
```diff
@@ -285,10 +285,9 @@ <h2>Instructions for submitting implementation reports</h2>
 <p>Reports should be submitted in Turtle format to
   <a href="mailto:public-rdf-star@w3.org">RDF-DEV CG</a> or via a Pull
   Request to the <a href="https://github.com/w3c/rdf-star/pulls">w3c/rdf-star</a>.
-  Add or update a file in the /releases directory named so that it can
-  be identified with your implementation.  See instructions on using the <a 
-  href="https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request">GitHub 
-  web interface</a> or submit a pull request from a forked repository using a branch other than &quot;main&quot;.</p>
+  Add or update a file in the /releases directory named so that it can be identified with your implementation.
```
TallTed, line 288:
```suggestion
  Add or update a file in the <code>/releases</code> directory named so that it can be identified with your implementation.
```

PR #147, opened 2021-03-31
Trig star

gkellogg:
Adds a section on TriG-star similar to N-Quads-star.

It seems that most of the work is already done by Turtle-star, but the discussion could go into more detail about processing, if anything thinks it would be useful.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/pull/147.html" title="Last updated on Apr 4, 2021, 7:22 PM UTC (6f7c3df)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/147/95b9762...6f7c3df.html" title="Last updated on Apr 4, 2021, 7:22 PM UTC (6f7c3df)">Diff</a>

PR #149, opened 2021-04-03
Split RDF-star and SPARQL-star manifests

afs:
Separate manifests for language and sparql tests (convenience for people who have different tools to run tests).


afs:
I tried to provide manifest-X.jsonld files but the existing Turtle example "tests/turtle/manifest.jsonld" does not work for me. I guess because of relative URIs in the worker manifests. 

This PR include the trig in #147 which has other manifest changes. I'll come back to the JSONLD when that's in.

gkellogg:
Note, we still have a single implementation report that reports on all tests, both SPARQL and concrete syntax. And the tests/index.html combines all the tests together.

Personally, I think this is appropriate, as the spec lists a single test suite and a single implementation report.

gkellogg:
The logic for make-jsonld-manifests needs a slight tweak to generate the JSON-LD versions of these manifests, which I can do as a commit to this PR.

gkellogg:
I rebased due to some conflicts.

afs:
Now we have some experience of the implementation reports process, is this still work doing?

Maybe it's better to have just the per-directory manifests - less duplication, maintenance.

In practice, I have not been running from any of the top-level `manifest*` files.

gkellogg:
We need the top-level HTML index, but that's built by hand. We could remove the manifest files in /tests, otherwise, but we should probably have them in sparql/, turtle/, and trig/ directories if we remove the top-level.

gkellogg (approved):

PR #154, opened 2021-04-09
Apache Jena EARL report - 2021-04-09

afs:


reports/jena-earl.ttl
```diff
@@ -0,0 +1,2058 @@
...
+        rdf:type         doap:Project ;
+        dc:creator       _:b0 ;
+        dc:title         "Apache Jena" ;
+        doap:homepage    "http://jena.apache.org/" ;
+        doap:maintainer  _:b0 ;
+        doap:developer   _:b0 ;
+        doap:name        "Apache Jena" ;
+        doap:release     [ rdf:type      doap:Version ;
+                           doap:created  "2021-04-01"^^xsd:date ;
+                           doap:name     "4.0.0"
```
gkellogg, line 2054:
```suggestion
                            doap:revision     "4.0.0"
```

gkellogg:
Looks good other then Version having a "revision" instead of a "name" (at least, that's how the reporting software interprets the data).

The "Apache Jena Community" should not try to link to a blank node, which is a small correction needed in the Haml template I'll take care of after the merge.

afs:
Fixed and tidied up.

afs:
Apologies - the version was wrong (the report generator is newer to conform with the submission requirements) than the system under test). There should not be a "SNAPSHOT". Fixed in the reports area.

gkellogg:
Do you want to submit an update, or patch in place?

afs:
I fixed it in-place (commit 2b791b06) and reran the report generator - a test of updating a report, not just adding one.

It updated and all went fine.

Issue #153, opened 2021-04-09
Issues running EARL report generator


afs:
Capturing points discovered while preparing a EARL PR:

1. The instructions say `/releases` but other reports are in `/reports`. But `/reports` is also the directory for the HTML production.
1. They are not linked to on the home page.
1. There are lot of stray red "X" on the reports. (these go away when there is more than one system in a section)
1. `doap:developer` may be a blank node.

pchampin:
the red "X"s indicate that we don't have two implementations passing that test (which is the success criterion on most RECs). In our case, we can probably disable this feature...

gkellogg:
> Capturing points discovered while preparing a EARL PR:
> 
> 1. The instructions say `/releases` but other reports are in `/reports`. But `/reports` is also the directory for the HTML production.

/reports is correct.

> 2. They are not linked to on the home page.

they should be, but more importantly, they’re linked from the spec.

> 3. There are lot of stray red "X" on the reports. (these go away when there is more than one system in a section)

what PA said.

> 4. `doap:developer` may be a blank node.

That should be okay.

pchampin:
@afs can we close this issue, or is there still a problem?

afs:
It's not a problem any longer.

The red 'X's were my misunderstanding.
The other points have been addressed.

gkellogg:
The red 'X's are implemented in the Haml template, as it was useful for WGs to know when there a feature is considered widely implemented. That doesn't apply for CGs, so we could easily update the template to not show the 'X's.

afs:
Now I understand their purpose, I'm fine with the status quo - I has assumed they marked an error in processing.

PR #152, opened 2021-04-07
Readme update

pchampin:
README.md now simply points to the homepage.

The implementation list has been converted to HTML, and is now pointed to by the homepage.

Adresses part of #150

PR #159, opened 2021-04-11
Link to implementation reports page

afs:
From #153.


gkellogg (commented):
I have a suggested change in #152 which completely changes the README.

afs:
This is index.html at https://w3.github.io/rdf-star/ . 

It adds one line for a link to implementation reports to go after the list of tests. (and reforms some HTML with no content change)

Long term, I can see the homepage changing to become about "explaining" but at the moment, it is an index to materials.

kinow (approved):

Issue #163, opened 2021-04-30
The SPARQL-star grammar does not cover nested triple syntax in expressions


Tpt:
The current SPARQL-star grammar does not seems to cover the [`sparql-star-expr-1`](https://w3c.github.io/rdf-star/tests/sparql/syntax/manifest.html#sparql-star-expr-1) test case:

```sparql
SELECT * {
  ?s ?p ?o .
  BIND(<<?s ?p ?o>> AS ?t)
}
```

Indeed, I do not see in the current drafpt any change of the [`Expression`](https://www.w3.org/TR/sparql11-query/#rExpression) grammar production of any of its nested productions. I believe the [`PrimaryExpression`](https://www.w3.org/TR/sparql11-query/#rPrimaryExpression) production should be updated to included RDF-star nest triples as well.

afs:
Thanks for the report. Indeed, `<< >>` should be in `PrimaryExpression`.

afs:
Done with PR #167

PR #167, opened 2021-05-01
SPARQL-star: Add embedded triple pattern to PrimaryExpression, define evaluation.

afs:
Addresses issue #163.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/pull/167.html" title="Last updated on May 6, 2021, 8:10 AM UTC (950082c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/167/ce2db5e...950082c.html" title="Last updated on May 6, 2021, 8:10 AM UTC (950082c)">Diff</a>

### DIFF
--- a/cg-spec/editors_draft.html
+++ b/cg-spec/editors_draft.html
@@ -895,6 +895,23 @@ SELECT ?claimer WHERE {
             <a data-cite="SPARQL11-QUERY#rTriplesNodePath">TriplesNodePath</a>
           </td>
         </tr>
+        <tr id="rPrimaryExpression">
+          <td>[119]</td>
+          <td>`PrimaryExpression`</td>
+          <td>::=</td>
+          <td>
+            <a data-cite="SPARQL11-QUERY#rBrackettedExpression">BrackettedExpression</a> `|` 
+            <a data-cite="SPARQL11-QUERY#rBuiltInCall">BuiltInCall</a> `|` 
+            <a data-cite="SPARQL11-QUERY#ririOrFunction">iriOrFunction</a> `|` 
+            <a data-cite="SPARQL11-QUERY#rRDFLiteral">RDFLiteral</a> `|` 
+            <a data-cite="SPARQL11-QUERY#rNumericLiteral">NumericLiteral</a> `|`
+            <a data-cite="SPARQL11-QUERY#rBooleanLiteral">BooleanLiteral</a> `|`
+            <a data-cite="SPARQL11-QUERY#rVar">Var</a>
+            <div style="font-weight: bold">
+            `|` <a href="#rExprEmbTP" style="font-weight: bold">ExprEmbTP</a>
+            </div>
+          </td>
+        </tr>
         <tr id="rBuiltInCall">
           <td>[121]</td>
           <td>`BuiltInCall`</td>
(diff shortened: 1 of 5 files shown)


Tpt:
I am not sure it is possible to reuse the `EmbTP ` rule here without restrictions: `EmbTP ` allows blank nodes that are not allowed in SPARQL expressions. What about introducing a new rule:

```
EmbTPExpr ::= '<<' PrimaryExpression PrimaryExpression PrimaryExpression '>>'
```

and make the produced value an alternative writing of the `TRIPLE` function?

afs:
Thanks for pointing that out.

PR set to "draft".

The idea is that it should be like a term in a graph pattern, so it is not full `PrimaryExpression` which allows `( expression )` and functions and built-ins. 

I think it is going to have to be written out as a specific form of vars or constants, and also `a` for the predicate slot. (And so avoids going near related matching forms SPARQL 1.1 does not have `?str@en` and `?lex^^:datatype`.)

For evaluation definition, `TRIPLE` can be used as it covers all the cases (and more).

afs:
Pushed an update for the grammar.

`ExprVarOrTerm` can be written using 177/DataValueTerm but that is currently specifically for VALUES or using `iriOrVar` but as rule names have implications so it seemed clearer to enumerate the cases.

It will need some stylistic reorganisation to fit with the grammar section.  Rules119a and 119b are placed close to use for clarity for now but style is to group new productions at the end.

cg-spec/editors_draft.html
```diff
@@ -895,6 +895,47 @@ <h2>Grammar</h2>
...
+            <a data-cite="SPARQL11-QUERY#rBooleanLiteral">BooleanLiteral</a> `|`
+            <a data-cite="SPARQL11-QUERY#rVar">Var</a>
+            <div style="font-weight: bold">
+            `|` <a href="#rExprEmbTP" style="font-weight: bold">ExprEmbTP</a>
+            </div>
+          </td>
+        </tr>
+        <tr id="rExprEmbTP">
+          <td>[119a]</td>
+          <td style="font-weight: bold">`ExprEmpTP`</td>
```
gkellogg, line 917:
```suggestion
          <td style="font-weight: bold">`ExprEmbTP`</td>
```

cg-spec/editors_draft.html
```diff
@@ -895,6 +895,47 @@ <h2>Grammar</h2>
...
+          <td>[119b]</td>
+          <td style="font-weight: bold">`ExprVarOrTerm`</td>
+          <td style="font-weight: bold">::=</td>
+          <td style="font-weight: bold">
+            <a data-cite="SPARQL11-QUERY#riri">iri</a> `|` 
+            <a data-cite="SPARQL11-QUERY#rRDFLiteral">RDFLiteral</a> `|` 
+            <a data-cite="SPARQL11-QUERY#rNumericLiteral">NumericLiteral</a> `|`
+            <a data-cite="SPARQL11-QUERY#rBooleanLiteral">BooleanLiteral</a> `|`
+            <a data-cite="SPARQL11-QUERY#rVar">Var</a>
+          </td>
```
gkellogg, line 937: Should this include `ExprVarOrTerm` in the same way that `EmbSubjectOrObject` includes `EmbTP`? Or, is that not possible in this context?

gkellogg (changes requested):
See my inline suggestions and questions.

Without adding `ExprEmbTP` to `ExprVarOrTerm` I can't parse one of our syntax tests.

cg-spec/editors_draft.html
afs: Thanks! Done.

cg-spec/editors_draft.html
afs: Done.

afs:
This PR now has:

* Grammar rules for `PrimaryExpression` involving `<<>>`.
* Two new bad syntax tests for `<<>> in expressions (the "good" case was already covered).
* Definition of evaluation of `<<>>` in an expression context (added new section 4.4.6). It is defined as being the same as `TRIPLE` with the expressions allowed by the syntax for `<<>>`.

gkellogg (approved):
This looks good; I have yet to try running your new tests, though.

afs:
@pchampin This now adds a section 4.4.6 for evaluation of `<< >>` in expressions.

pchampin (approved):

Issue #165, opened 2021-04-30
create "outstanding issues" in the CG-report as being the things to complete before we're done


pchampin:
assigned during today's called: https://w3c.github.io/rdf-star/Minutes/2021-04-30.html#a02

pchampin:
This was discussed during the call on 2021-05-14: https://w3c.github.io/rdf-star/Minutes/2021-05-14.html#t02
Summary:
I consider that we can have a "first final report" when
* #121 is solved
* the intro section is tidied up

Other issues can, IMHO, be addressed afterward (and further public drafts / "final" reports can still be published afterwards).

Issue #94, opened 2021-02-12
SPARQL-star update example in the overview/primer part


hartig:
The overview/primer part of the report should also contain an example that hints at the fact that the spec also covers updates.

hartig:
This issue may be addressed by simply moving the informal-overview section that PR #98 adds to Section 5 into the general overview/primer part.

pchampin:
@hartig I didn't follow your suggestion above, and created a new example for §1. I thought that the informal description in §5 was too detailed for the intro (but of course, it stays as is in §5).

PR #171, opened 2021-05-20
Tidy up intro

pchampin:
addresses #94 and #165

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/pull/171.html" title="Last updated on May 28, 2021, 11:38 AM UTC (e5abb6b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/171/be79326...e5abb6b.html" title="Last updated on May 28, 2021, 11:38 AM UTC (e5abb6b)">Diff</a>

### DIFF
--- a/cg-spec/editors_draft.html
+++ b/cg-spec/editors_draft.html
@@ -153,7 +153,7 @@
 </head>
 <body>
   <section id="abstract">
-    <p>The Resource Description Framework (RDF) is a general-purpose framework for representing information on the Web. RDF-star extends RDF with a convenient way to make statements about other statements. This specification defines the abstract syntax of RDF-star as an extension of RDF's, it extends a number of RDF concrete syntaxes to support the new abstract syntax, as well as RDF's formal semantics. This specification also extends the SPARQL language to allow querying and updating of RDF-star data.</p>
+    <p>The Resource Description Framework (RDF) is a general-purpose framework for representing information on the Web. RDF-star extends RDF with a convenient way to make statements about other statements. This specification defines the abstract syntax of RDF-star as an extension of RDF's. It extends a number of RDF concrete syntaxes to support the new abstract syntax. It also extends RDF's formal semantics. Finally, this specification extends the SPARQL language to allow querying and updating of RDF-star data.</p>
   </section>
   <section id="sotd">
   </section>
@@ -164,10 +164,10 @@
     <section class="informative">
       <h2>Background and Motivation</h2>
 
-      <p>The RDF data model lets you state facts in three-part subject-predicate-object statements known as triples. For example, with a single RDF triple you can say that employee38 has a familyName of "Smith". A triple's predicate is a property specified with an IRI (an Internationalized version of a URI), identifying that property in a globally unambiguous way. A triple's subject and object can each be an IRI referencing any entity, and the object can also be a literal value such as "Smith" or data of other types such as dates, numbers, or Boolean values.</p>
+      <p>The RDF data model lets you state facts in three-part subject-predicate-object statements known as triples. For example, with a single RDF triple, you can say that employee38 has a familyName of "Smith". A triple's predicate is a property specified with an IRI (an Internationalized version of a URI), identifying that property in a globally unambiguous way. A triple's subject and object can each be an IRI referencing any entity, and the object can also be a literal value such as "Smith" or data of other types such as dates, numbers, or Boolean values.</p>
 
       <p>Sometimes, we want the subject or object of a triple to refer to another triple. For example, the statement "according to employee22, employee38 has a jobTitle of 'Assistant Designer'" can be modeled as a triple with "according to" as its predicate, employee22 as its object, and another triple as its subject, namely the triple "employee38 has a jobTitle of 'Assistant Designer'". This use of a triple as the subject or object resource of another triple so that we can say things about that triple is known as <a href='https://www.w3.org/TR/rdf11-mt/#reification'>reification</a>.
-      The concept of reification has always been part of RDF, but expressing it in RDF concrete syntaxes such as Turtle, N-Triples, and RDF/XML, as well as processing or querying it with SPARQL, has been verbose and cumbersome [[RDF-STAR-FOUNDATION]].</p>
+      The concept of reification has always been part of RDF, but expressing it in most RDF concrete syntaxes such as Turtle and N-Triples, as well as processing or querying it with SPARQL, has been verbose and cumbersome [[RDF-STAR-FOUNDATION]].</p>
 
     </section>
 
--- a/cg-spec/editors_draft.html
+++ b/cg-spec/editors_draft.html
@@ -153,7 +153,7 @@
 </head>
 <body>
   <section id="abstract">
-    <p>TODO</p>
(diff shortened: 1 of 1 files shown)


cg-spec/editors_draft.html
```diff
@@ -153,7 +153,7 @@
 </head>
 <body>
   <section id="abstract">
-    <p>TODO</p>
+    <p>The Resource Description Framework (RDF) is a general-purpose framework for representing information on the Web. RDF-star extends RDF with a convenient way to make statements about other statements. This specification defines the abstract syntax of RDF-star as an extension of RDF's, it extends a number of RDF concrete syntaxes to support the new abstract syntax, as well as RDF's formal semantics. This specification also extends the SPARQL language to allow querying and updating of RDF-star data.</p>
```
TallTed, line 156:
```suggestion
    <p>The Resource Description Framework (RDF) is a general-purpose framework for representing information on the Web. RDF-star extends RDF with a convenient way to make statements about other statements. This specification defines the abstract syntax of RDF-star as an extension of RDF's. It extends a number of RDF concrete syntaxes to support both the new abstract syntax and RDF's formal semantics. This specification also extends the SPARQL language to allow querying and updating of RDF-star data.</p>
```

cg-spec/editors_draft.html
```diff
@@ -164,19 +164,18 @@ <h1>Introduction</h1>
     <section class="informative">
       <h2>Background and Motivation</h2>
 
-      <p>TODO, citing [[RDF-STAR-FOUNDATION]]</p>
+      <p>The RDF data model lets you state facts in three-part subject-predicate-object statements known as triples. For example, with a single RDF triple you can say that employee38 has a familyName of "Smith". A triple's predicate is a property specified with an IRI (an Internationalized version of a URI), identifying that property in a globally unambiguous way. A triple's subject and object can each be an IRI referencing any entity, and the object can also be a literal value such as "Smith" or data of other types such as dates, numbers, or Boolean values.</p>
```
TallTed, line 167:
```suggestion
      <p>The RDF data model lets you state facts in three-part subject-predicate-object statements known as triples. For example, with a single RDF triple, you can say that employee38 has a familyName of "Smith". A triple's predicate is a property specified with an IRI (an Internationalized version of a URI), identifying that property in a globally unambiguous way. A triple's subject and object can each be an IRI referencing any entity, and the object can also be a literal value such as "Smith" or data of other types such as dates, numbers, or Boolean values.</p>
```

TallTed (changes requested):

rat10:
[sorry for not being able to suggest the following change via  proper GIT commands!]

In section 1.2, at the end of the second paragraph, I propose to replace 

`The concept of reification has always been part of RDF, but expressing it in RDF concrete syntaxes such as Turtle, N-Triples, and RDF/XML, as well as processing or querying it with SPARQL, has been verbose and cumbersome.`

with 

`The concept of reification has always been part of RDF, but expressing it in RDF concrete syntaxes such as Turtle and N-Triples, as well as processing or querying it with SPARQL, has been verbose and cumbersome. RDF/XML provides syntactic sugar for reification by means of an :ID attribute but for other reasons has become mostly a legacy format.`

As Jerven Bollmann mentioned, the rdf:ID attribute is the reason why they still use RDF/XML in Uniprot.

cg-spec/editors_draft.html
```diff
@@ -164,19 +164,18 @@ <h1>Introduction</h1>
...
+
+      <p>Sometimes, we want the subject or object of a triple to refer to another triple. For example, the statement "according to employee22, employee38 has a jobTitle of 'Assistant Designer'" can be modeled as a triple with "according to" as its predicate, employee22 as its object, and another triple as its subject, namely the triple "employee38 has a jobTitle of 'Assistant Designer'". This use of a triple as the subject or object resource of another triple so that we can say things about that triple is known as <a href='https://www.w3.org/TR/rdf11-mt/#reification'>reification</a>.
+      The concept of reification has always been part of RDF, but expressing it in RDF concrete syntaxes such as Turtle, N-Triples, and RDF/XML, as well as processing or querying it with SPARQL, has been verbose and cumbersome [[RDF-STAR-FOUNDATION]].</p>
```
hartig, line 170: Here is the suggestion made by @rat10 in his [recent comment](https://github.com/w3c/rdf-star/pull/171#issuecomment-849140936) captured directly as a suggested change to this PR:

```suggestion
      The concept of reification has always been part of RDF [[RDF-STAR-FOUNDATION]], but expressing it in RDF concrete syntaxes such as Turtle and N-Triples, as well as processing or querying it with SPARQL, has been verbose and cumbersome. RDF/XML provides syntactic sugar for reification by means of an :ID attribute but, for other reasons, has become mostly a legacy format.</p>
```

cg-spec/editors_draft.html
pchampin: That's a misinterpretation of what I meant. "as well as" was about "it extends", not about "to support"... I'll rephrase that with shorter sentences (as I believe this is the essence of this suggestion), hopefully in a less ambiguous way.

cg-spec/editors_draft.html
pchampin: I agree to remove RDF/XML from this list, because indeed reification in this syntax is not as verbose as in others.

However, calling it a legacy format may sound to harsh to some people. It may also give the impression that an extension of RDF/XML to support RDF-star would be out of scope, which 1) is not the case, and 2) would [worry some people](https://lists.w3.org/Archives/Public/public-rdf-star/2021May/0079.html).

So I would rather just remove RDF/XML from the list, and may be add "most" in front of "concrete syntaxes" (

afs: > `but for other reasons has become mostly a legacy format.`

Remove this. Unrelated and opinion.

TallTed (approved):

gkellogg (approved):

cg-spec/editors_draft.html
pchampin: It was, in e5abb6b

rat10:
I would have been fine with removing "but for other reasons has become mostly a legacy format" but not with removing the reference to RDF/XML itself. With that sub sentence I didn't want to hurt anyones feelings but provide some sort of reference to the fact that RDF/XML is not anymore the go-to format for all things RDF that it has once been, so it's practical importance is limited nowadays. But as I said, I'm fine with removing any such statement or rewording it in the nicest possible way.

What is important however is the fact that RDF/XML is the only standardized syntax that provides syntactic sugar for RDF reification in form of the ID attribute. Providing a less cumbersome alternative to the verbose standard reification quadlet was not too long ago - before the proposed semantics was introduced - the central value proposition of RDF-star. So RDF/XML is of very high relevance in the context of RDF-star, and much higher than all the other syntaxes mentioned.

Also one of the most concrete use cases ilisted for RDF-star is that of UniProt which would like to get away from RDF/XML but makes heavy use of the ID attribute to sweeten reification syntactically. UniProt, as expressed by Jerven Bolleman would like to switch to RDF-star but would also need its model-theoretic semantics to support referential transparency.

So this is indeed an interesting topic with very relevant aspects and deleting it from the report entirely because a sub sentence might hurt somebodies feelings is a very debatable editorial decision. RDF/XML is now only mentioned once, in section 3.5 "Other Concrete Syntaxes" in the context of extendendig those to support RDF-star. 

In academic papers an important section is always that of "Related Work". I don't know fro certain what the customs are w.r.t. W3C Recommendations or Community Reports but I would expect some of the same attitude towards openness and constructive discussion and dialogue. That certainly includes mentioning the closest related approach to the problem.

Issue #166, opened 2021-04-30
add a section about the rationale for semantic opacity / transparency


pchampin:
assigned during today's call: https://w3c.github.io/rdf-star/Minutes/2021-04-30.html#a03

pchampin:
Closing, as PR #173 addresses this issue.

PR #173, opened 2021-05-20
Discussion on referential opacity

pchampin:
addresses #166

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/pull/173.html" title="Last updated on Jun 4, 2021, 6:24 AM UTC (dccae50)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/173/5b20a3f...dccae50.html" title="Last updated on Jun 4, 2021, 6:24 AM UTC (dccae50)">Diff</a>

### DIFF
--- a/cg-spec/editors_draft.html
+++ b/cg-spec/editors_draft.html
@@ -2548,7 +2548,7 @@ style="color:black;font-weight:bold">sameTerm</code></dfn></b>
       <p>We define a mapping |L| that maps any <a>IRI</a> or <a>literal</a> |t| to a literal with:</p>
       <ul>
         <li>`xsd:string` as its <a>datatype</a>, and</li>
-        <li>the <a data-cite="N-TRIPLES#canonical-ntriples">canonical N-Triples</a> representation of |t| as its <a>lexical form</a> [[N-TRIPLES]]. If |t| is itself a literal with the `xsd:string` <a>datatype</a>, the representation MUST simply match the <a href="https://www.w3.org/TR/n-triples/#grammar-production-STRING_LITERAL_QUOTE">`STRING_LITERAL_QUOTE`</a> production (i.e., without an explicit datatype IRI).</li>
+        <li>the <a data-cite="N-TRIPLES#canonical-ntriples">canonical N-Triples</a> representation of |t| as its <a>lexical form</a> [[N-TRIPLES]]. If |t| is itself a literal with the `xsd:string` <a>datatype</a>, the representation MUST simply match the <a href="https://www.w3.org/TR/n-triples/#grammar-production-STRING_LITERAL_QUOTE">`STRING_LITERAL_QUOTE`</a> production (i.e., without an explicit <a>datatype</a> <a>IRI</a>).</li>
       </ul>
 
       <p>Given an <a>RDF-star graph</a> |G|, the following steps transform it into an <a>RDF graph</a> that we call |unstar|(|G|).</p>
@@ -2563,13 +2563,13 @@ style="color:black;font-weight:bold">sameTerm</code></dfn></b>
           <li>Pick an <a>RDF-star triple</a> (|s|, |p|, |o|) in the <a>constituents</a> of |G| such that neither |s| nor |o| is an <a>embedded triple</a>.</li>
           <li>Mint a fresh blank node |b| (i.e., such that |b| is not in the <a>constituents</a> of |G|).</li>
           <li>Replace with |b| all occurrences of (|s|, |p|, |o|) in the <a>subject</a> or <a>object</a> position of an <a>asserted</a> or <a>embedded triple</a> of |G|.</li>
-          <li>Add the following <a>asserted triples</a> to |G|:<ul>
+          <li id="unstar-add-triples">Add the following <a>asserted triples</a> to |G|:<ul>
             <li>(|b|, `unstar:subject`,`   `|s|)</li>
             <li>(|b|, `unstar:predicate`,` `|p|)</li>
             <li>(|b|, `unstar:object`,`    `|o|) unless |o| is an <a>ill-typed</a> literal</li>
-            <li>(|b|, `unstar:subjectLexical`,`   `|L|(|s|)) unless |s| is <a>blank node</a></li>
+            <li>(|b|, `unstar:subjectLexical`,`   `|L|(|s|)) unless |s| is a <a>blank node</a></li>
             <li>(|b|, `unstar:predicateLexical`,` `|L|(|p|))</li>
-            <li>(|b|, `unstar:objectLexical`,`    `|L|(|o|)) unless |o| is <a>blank node</a></li>
+            <li>(|b|, `unstar:objectLexical`,`    `|L|(|o|)) unless |o| is a <a>blank node</a></li>
           </ul></li>
         </ol></li>
       </ol>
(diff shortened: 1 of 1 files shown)


cg-spec/editors_draft.html
```diff
@@ -2341,9 +2340,118 @@ <h2>Other uses of the |unstar| mapping</h2>
 
       </section>
 
+      <section>
+        <h2>Referential opacity</h2>
+
+        <p>The way the |unstar| mapping is defined, the denotation of an embedded triple does not only depend on the denotation of its component terms, but also on their syntactical form (for IRIs and literals). This behavior is referred to as <a href="https://en.wikipedia.org/wiki/Opaque_context">referential opacity</a>. A consequence is that two different triples, that would be semantically equivalent when asserted, are not automatically considered to denote the same thing when embedded. In the example below, assuming <a data-cite="RDF11-MT#datatype-entailment">D-entailment</a>: while `"42"^^xsd:integer` and `"042"^^xsd:integer` denote the same thing (namely, the number 42), the two embedded triples are allowed to have different denotations, and so the entailment does not hold.</p>
```
TallTed, line 2346:
```suggestion
        <p>The way the |unstar| mapping is defined, the denotation of an embedded triple does not only depend on the denotation of its component terms, but also on their syntactical form (for IRIs and literals). This behavior is referred to as <a href="https://en.wikipedia.org/wiki/Opaque_context">referential opacity</a>. A consequence is that two different triples, that would be semantically equivalent if asserted, are not automatically considered to denote the same thing when embedded. In the example below, assuming <a data-cite="RDF11-MT#datatype-entailment">D-entailment</a>: while `"42"^^xsd:integer` and `"042"^^xsd:integer` denote the same thing (namely, the number 42), the two embedded triples are allowed to have different denotations, and so the entailment does not hold.</p>
```

cg-spec/editors_draft.html
```diff
@@ -2341,9 +2340,118 @@ <h2>Other uses of the |unstar| mapping</h2>
...
+
+        <h2g2.ttl> :contains << :lue :answer "42"^^xsd::integer >>.
+
+        #### does NOT entail
+
+        <h2g2.ttl> :contains << :lue :answer "042"^^xsd::integer >>.
+        -->
+        </pre>
+
+        <p>Similarly, in the example below and under <a href="https://www.w3.org/TR/owl2-overview/">OWL entailment</a> [[OWL2-Overview]]: while `:superman` and `:clark` are inferred to denote one and the same person (because of the `owl:sameAs` relationship between them), the two embedded triples are not automatically considered to denote the same thing.</p>
```
TallTed, line 2364:
```suggestion
        <p>Similarly, in the example below and under <a href="https://www.w3.org/TR/owl2-overview/">OWL entailment</a> [[OWL2-Overview]], while `:superman` and `:clark` are inferred to denote one and the same person (because of the `owl:sameAs` relationship between them), the two embedded triples are not automatically considered to denote the same thing.</p>
```

cg-spec/editors_draft.html
```diff
@@ -2341,9 +2340,118 @@ <h2>Other uses of the |unstar| mapping</h2>
...
+        </pre>
+
+        <p>Note that the current semantics does not forbid the additional inferences to be drawn. However, they would require additional knowledge and an appropriate <a data-cite="RDF11-MT#dfn-semantic-extension">semantic extension</a>, not specified in this document. This makes RDF-star less readily usable for (but not incompatible with) that latter kind of use-cases.</p>
+
+      </section>
+
+      <section>
+        <h2>Alternatives to referential opacity</h2>
+
+        <p>Given the potential drawbacks, described in the previous section, of the referential opacity provided by the current semantics, the group has considered alternative semantics. No clear consensus was reached, however, to decice whether these altenatives should be used as a replacement, or as a <a data-cite="RDF11-MT#dfn-semantic-extension">semantic extension</a>, of the current semantics.</p>
```
TallTed, line 2419:
```suggestion
        <p>Given the potential drawbacks of the referential opacity provided by the current semantics, as described in the previous section, the group has considered alternative semantics. No clear consensus was reached, however, to decide whether these alternatives should be used as a replacement, or as a <a data-cite="RDF11-MT#dfn-semantic-extension">semantic extension</a>, of the current semantics.</p>
```

cg-spec/editors_draft.html
```diff
@@ -2341,9 +2340,118 @@ <h2>Other uses of the |unstar| mapping</h2>
...
+              </ul>
+          </li>
+        </ul>
+
+      </section>
+
     </section>
 
-    <div class=note>Multiple design choices have been discussed by the community group for the semantics, that have been <a href="https://github.com/w3c/rdf-star/issues/95">summarized in a Github issue</a>.</div>
+    <div class=note>Other design choices have also been discussed by the community group for the semantics, that have been <a href="https://github.com/w3c/rdf-star/issues/95">summarized in a Github issue</a>.</div>
```
TallTed, line 2454:
```suggestion
    <div class=note>Other design choices for the semantics have also been discussed by the community group; these have been <a href="https://github.com/w3c/rdf-star/issues/95">summarized in a Github issue</a>.</div>
```

rat10:
It turns things upside down to call reinstantiating referential transparency of embedded triples a 'semantic extension'.

hartig:
> It turns things upside down to call reinstantiating referential transparency of embedded triples a 'semantic extension'.

@rat10 care to elaborate more on this statement?

rat10:
> > It turns things upside down to call reinstantiating referential transparency of embedded triples a 'semantic extension'.
> 
> @rat10 care to elaborate more on this statement?
All of RDF operates on the level of interpretation, only embedded triples per the proposed semantics are treated as syntactic entities. This prevents entailments that would otherwise be normal. So the proposed semantics should rather be called a "semantic reduction" than the re-instatement of referential transparency a semantic extension.

pchampin:
@rat10 The text in 6.4.5 is considering two options:
* A: replace the opaque entailment (OE) by an alternative transparent entailment (TE)
* B: keep both OA and TA

IF option B is chosen, THEN TE is a semantic extension of OE, per [the definition of 'semantic extension' in the spec](https://www.w3.org/TR/rdf11-mt/#dfn-semantic-extension).

You might consider B to be a bad idea, but that does not make the implication above false, does it?

PS: Maybe your argument was not so much about the definition of 'semantic extension' than about the connotation it may carry. However, the fact that the term links to its definition makes it clear, I think, that it is to be understood as meaning no more and no less than said definition. There is no judgement value in the the word "extension".

pchampin:
> @pchampin The spec you cite speaks of semantic extensions as vocabularies 

Does it? I read "A particular such set of semantic assumptions is called a **semantic extension**." In which dictionnary is "vocabulary" synonymous with "a set of semantic assumptions"??

Interpreting spec text in such creative ways is not helping either.

> that when applied may make a given RDF graph "support more extensive entailments than are sanctioned by the basic RDF semantics". Neither is RDF-star a vocabulary nor does it lead to more entailments, rather to the contrary.

If by "the contrary" you mean "it leads to less entailments", then this is simply wrong. Any entailment that is true in RDF is also true RDF-star.

But apart from that, what are you trying to demonstrate here? That RDF-star can not be called a semantic extension of RDF? Very well, I agree with that, and I don't see what in this PR can be interpreted as making that claim. What the text says is that the alternative (ref. transparent) semantics could be defined as a semantic extension of the current (ref. opaque) semantics.

Yes, connotations and subtext are important. But the text is important too!

cg-spec/editors_draft.html
```diff
@@ -2660,9 +2659,118 @@ <h2>Other uses of the |unstar| mapping</h2>
 
       </section>
 
+      <section>
+        <h2>Referential opacity</h2>
+
+        <p>The way the |unstar| mapping is defined, the denotation of an embedded triple does not only depend on the denotation of its component terms, but also on their syntactical form (for IRIs and literals). This behavior is referred to as <a href="https://en.wikipedia.org/wiki/Opaque_context">referential opacity</a>. A consequence is that two different triples, that would be semantically equivalent if asserted, are not automatically considered to denote the same thing when embedded. In the example below, assuming <a data-cite="RDF11-MT#datatype-entailment">D-entailment</a>: while `"42"^^xsd:integer` and `"042"^^xsd:integer` denote the same thing (namely, the number 42), the two embedded triples are allowed to have different denotations, and so the entailment does not hold.</p>
```
gkellogg, line 2665: Instead of |unstar|, this should probably be a reference to a defined term _unstar mapping_, which can be defined in section 6.1. Otherwise, the notion of unstar mapping is never really defined.

cg-spec/editors_draft.html
```diff
@@ -2660,9 +2659,118 @@ <h2>Other uses of the |unstar| mapping</h2>
...
+        :superman owl:sameAs :clark.
+        << :superman :can :fly >> :reportedBy :clark.
+
+        #### does NOT entail
+
+        << :clark :can :fly >> :reportedBy :clark.
+        -->
+        </pre>
+
+        <p>This is well suited for use-cases where the syntactical form of the annotated statements may be significant (like <a href="../UCR/rdf-star-ucr.html#annotate-statements-from-other-named-graphs-and-datasets">annotating statements from other graphs</a> or <a href="../UCR/rdf-star-ucr.html#annotate-commit-deltas">annotating commit deltas</a>). Other use-cases, on the other hand, would expect some degree of referential transparency. Consider the case of <a href="../UCR/rdf-star-ucr.html#meta-properties-over-properties">attributed/evidenced triples</a>. The annotation is understood to be about the fact asserted by the triple rather than the triple itself. As illustrated in the example below, one might then expect more inferences than provided by the current semantics.</p>
```
gkellogg, line 2702:
```suggestion
        <p>This is well suited for use-cases where the syntactical form of the annotated statements may be significant (like <a href="../UCR/rdf-star-ucr.html#annotate-statements-from-other-named-graphs-and-datasets">annotating statements from other graphs</a> or <a href="../UCR/rdf-star-ucr.html#annotate-commit-deltas">annotating commit deltas</a>). On the other hand, other use-cases would expect some degree of referential transparency. Consider the case of <a href="../UCR/rdf-star-ucr.html#meta-properties-over-properties">attributed/evidenced triples</a>. The annotation is understood to be about the fact asserted by the triple rather than the triple itself. As illustrated in the example below, one might then expect more inferences than provided by the current semantics.</p>
```

gkellogg (changes requested):
I'm happy with the interpretation, just some minor suggestions on grammar and linking.

afs (approved):

cg-spec/editors_draft.html
pchampin: Good idea. Done in 5930b88

gkellogg (approved):

cg-spec/editors_draft.html
```diff
@@ -2631,29 +2631,29 @@ <h2>Remarks</h2>
...
+        <p>It is important to notice that the definitions given <a href="#entailment-of-rdf-star-graphs">above</a> for <a>satisfiability</a> and <a>entailment</a> of <a>RDF-star graphs</a> do not mandate that implementations actually transform <a>RDF-star graphs</a> using the <a>unstar mapping</a> defined above. It should be possible to implement the semantics by directly handling RDF-star's <a href="#concepts">abstract syntax</a>.</p>
 
       </section>
 
       <section>
         <h2>Combining RDF-star graphs</h2>
 
-        <p>For implementations that rely on the |unstar| mapping defined <a href="#mapping">above</a>, care must be taken when the resulting <a>RDF graphs</a> are combined through union or <a>merge</a>. Given two <a>RDF-star graphs</a> |G| and |H|, it may be the case that |unstar|(|G| ∪ |H|) ≠ |unstar|(|G|) ∪ |unstar|(|H|).</p>
+        <p>For implementations that rely on the <a>unstar mapping</a>, care must be taken when the resulting <a>RDF graphs</a> are combined through union or <a>merge</a>. Given two <a>RDF-star graphs</a> |G| and |H|, it may be the case that <a>unstar</a>(|G| ∪ |H|) ≠ <a>unstar</a>(|G|) ∪ <a>unstar</a>(|H|).</p>
```
TallTed, line 2641:
```suggestion
        <p>Care must be taken when implementations that rely on the <a>unstar mapping</a>, combine the resulting <a>RDF graphs</a> through union or <a>merge</a>. Given two <a>RDF-star graphs</a> |G| and |H|, it may be the case that <a>unstar</a>(|G| ∪ |H|) ≠ <a>unstar</a>(|G|) ∪ <a>unstar</a>(|H|).</p>
```

TallTed (changes requested):
Just a bit of text masssage

rat10:
> > @pchampin The spec you cite speaks of semantic extensions as vocabularies
> 
> Does it? I read "A particular such set of semantic assumptions is called a **semantic extension**." In which dictionnary is "vocabulary" synonymous with "a set of semantic assumptions"??

I read before that: "Also, particular IRI vocabularies may be given meanings by other specifications or conventions. When such extra meanings are assumed, a given RDF graph may support more extensive entailments than are sanctioned by the basic RDF semantics. In general, the more assumptions that are made about the meanings of IRIs in an RDF graph, the more entailments follow from those assumptions." So particular vocabularies can be given extra meanings and assuming those extra meanings leads to a semantic extension. 

> Interpreting spec text in such creative ways is not helping either.

So I was imprecise but not very creative.

> > that when applied may make a given RDF graph "support more extensive entailments than are sanctioned by the basic RDF semantics". Neither is RDF-star a vocabulary nor does it lead to more entailments, rather to the contrary.

Yes, here I was clearly tangling things up: RDF-star is neither a vocabulary nor a semantic extension.
 
> If by "the contrary" you mean "it leads to less entailments", then this is simply wrong. Any entailment that is true in RDF is also true RDF-star.

And here I beg to differ. How this argument turns out depends very much on how one defines the setting. 
If one defines RDF-star as the new normal then of course adding more entailments by interpreting embedded triples as referentially transparent extends the set of possible entailments. If that is controlled through some vocabulary then it is a semantic extension.
If one however looks at RDF before and after the introduction of referential opacity for embedded triples, one sees that embedded triples don't drive entailments as one would expect them to - given the firm foundation of RDF semantics in referential transparency (and that not only as a mere side effect or coincidence but as an important and fundamental design decision). In this perspective RDF-star is very much restricting entailments in RDF. The propsed semantics restricts possible entailments from embedded triples in ways that are very surprising even to seasoned semanticists because such restrictions (or "prudence" as you prefer to call it) AFAICT are applied nowhere else in RDF except in Notation 3 formulas. 
 
> But apart from that, what are you trying to demonstrate here? That RDF-star can not be called a semantic extension of RDF? Very well, I agree with that, and I don't see what in this PR can be interpreted as making that claim. What the text says is that the alternative (ref. transparent) semantics could be defined as a semantic extension of the current (ref. opaque) semantics.
> 
> Yes, connotations and subtext are important. But the text is important too!

The report IMO needs to be extended by a self-contained discussion of the merits and shortcomings of referential opacity of embedded triples - before it becomes "final" for the first time. A better place might however be section 2.2. That would concentrate discussions of the semantics on a more fundamental level in one place and technical details like how to get from opaque to transparent semantics wouldn't need to set the scene right. Maybe make that a new issue?

pchampin:
This was discussed during yesterday's call: https://w3c.github.io/rdf-star/Minutes/2021-06-04.html#t04

pchampin:
@rat10
> The report IMO needs to be extended by a self-contained discussion of the merits and shortcomings of referential opacity of embedded triples

That's indeed the purpose of this PR, and the role that §6.4.4 and §6.4.5 aim to play. As agreed during the last call, feel free to propose changes if you feel that they don't play that role appropriately.

> > Any entailment that is true in RDF is also true RDF-star.

> How this argument turns out depends very much on how one defines the setting. 

Every term in my claim above has a precise definition, either in the RDF spec or in our CG report. So the setting seems pretty unambiguous to me.

> And here I beg to differ. 

Let me rephrase the assertion above: given any two RDF graphs G1 and G2, if G1 entails G2 in simple RDF entailment, then G1 entails G2 in the newly defined RDF-star entailment. I think this trivially follows from the way RDF-star entailment is defined. Or do you have a counter example?

> (...) embedded triples don't drive entailments as one would expect them to.

RDF does not have any notion of embedded triples. As such, any *expectation* one might have, based on the  RDF standard, about embedded triples, is an extrapolation, and as such, partly subjective. You are absolutely entitled to make such extrapolations, but you should not make them sound like they are necessarily shared by everyone, or like they are an integral part of the RDF standard.

rat10:
@pchampin 
> @rat10
> 
> > The report IMO needs to be extended by a self-contained discussion of the merits and shortcomings of referential opacity of embedded triples
> 
> That's indeed the purpose of this PR, and the role that §6.4.4 and §6.4.5 aim to play. As agreed during the last call, feel free to propose changes if you feel that they don't play that role appropriately.

Working on it...

> > > Any entailment that is true in RDF is also true RDF-star.
> 
> > How this argument turns out depends very much on how one defines the setting.
> 
> Every term in my claim above has a precise definition, either in the RDF spec or in our CG report. So the setting seems pretty unambiguous to me.

Your setting is unambiguous, but its not the only relevant one. Maybe you fail to imagine a setting where RDF-star is not yet proposed and in which referential transparency is the norm? That is the world we live in and in that world, or setting, the proposed semantics for RDF-star suppresses entailments in a surprising way.
 
> > And here I beg to differ.
> 
> Let me rephrase the assertion above: given any two RDF graphs G1 and G2, if G1 entails G2 in simple RDF entailment, then G1 entails G2 in the newly defined RDF-star entailment. I think this trivially follows from the way RDF-star entailment is defined. Or do you have a counter example?
> 
> > (...) embedded triples don't drive entailments as one would expect them to.
> 
> RDF does not have any notion of embedded triples. As such, any _expectation_ one might have, based on the RDF standard, about embedded triples, is an extrapolation, and as such, partly subjective. You are absolutely entitled to make such extrapolations, but you should not make them sound like they are necessarily shared by everyone, or like they are an integral part of the RDF standard.

AFIACT the concept of referential opacity is not part of the RDF 1.1 semantics. The closest thing to RDF-star in RDF, standard reification, is explicitly referentially transparent and RDF-star not long ago was advertized as being syntactic sugar for RDF standard reification. The ID attribute in RDF/XML as the only standardized syntactic sugar for referring to other statements in RDF is based on standard reification in RDF and therefore referentially transparent. Whoever stopped following the development of RDF-star last automn would be perfectly excused for expecting a referentially transparent semantics. So please explain to me what is so overly subjective about my "extrapolation". 

To the contrary the proposed semantics is optimized to enable use cases that have so far not been possible at all in RDF - explainable AI, to the letter accurate representations of statements - and also not on any popular wish list that I'm aware of (or do you e.g. find them in the [Easier RDF](https://github.com/w3c/EasierRDF) wiki?). To achieve that goal it suppresses co-denotations and entailments that everywhere on the semantic web are taken for granted and sometimes even essential. How can that semantics be called anything less than unexpected?

Issue #175, opened 2021-05-28
add ref to https://www.w3.org/TR/dx-prof-conneg/ (with warning about SPARQL protocol)


pchampin:
per https://w3c.github.io/rdf-star/Minutes/2021-05-21.html#a03

pchampin:
closing, as the discussion will now occur in PR #177.

PR #177, opened 2021-06-03
Add discussion of TR 'Content Negotiation by Profile'

afs:
This address #175.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/pull/177.html" title="Last updated on Jun 4, 2021, 4:35 PM UTC (a1d1641)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/177/5b20a3f...a1d1641.html" title="Last updated on Jun 4, 2021, 4:35 PM UTC (a1d1641)">Diff</a>


pchampin (changes requested):
LGTM, apart from the `x.` vs `x-` distinction in RFC6838

cg-spec/editors_draft.html
```diff
@@ -2671,12 +2671,36 @@ <h2>IANA Considerations</h2>
...
 
-    <p>In the meantime, implementers of RDF-star are faced with a dilemma: use the standard media-type identifiers as if their definition had been already updated (optimistic approach); or make it explicit that they are prone to produce or consume embedded triples (pessimistic approach), perhaps by using a different temporary media-type from the unregistered `x.` tree [[RFC6838]], profile links [[RFC6906]], or custom HTTP headers.</p>
+    <p>In the meantime, implementers of RDF-star are faced with a dilemma: use the standard media-type identifiers as if their definition had been already updated (optimistic approach); or make it explicit that they are prone to produce or consume embedded triples (pessimistic approach), perhaps by using a different temporary media-type from the unregistered `x-` tree [[RFC6838]], profile links [[RFC6906]], or custom HTTP headers.</p>
```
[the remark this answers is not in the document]
afs: Happy to revert the change.

My understanding is that `x-` was an old convention for "experimental, temporary" 
(and now completely not a good idea ([Grandfathered Media Types](https://datatracker.ietf.org/doc/html/rfc6838#appendix-A))

`x.` is space of unregistered named for limited scope types "in private, local environments" which are not temporary, so like more "external" but specifically not "temporary".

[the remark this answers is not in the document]
pchampin: as you write, `x-` is an old convention; and any temporary media-type will not be registered (!) so by nature it belongs to the "unregistered" namespace, right?
Although of course we know that "temporary-ness" is an illusion in that space.

gkellogg (approved):

cg-spec/editors_draft.html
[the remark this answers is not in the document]
TallTed: "in private, local environments" says, to me, that the `x.` tree is not appropriate, either -- because we're talking about *global*, *experimental*, and (hopefully) *temporary* (until a long-term media type is determined) deployment.

cg-spec/editors_draft.html
```diff
@@ -2671,12 +2671,36 @@ <h2>IANA Considerations</h2>
...
+      <a data-cite="DX-PROF-CONNEG#introduction">Content Negotiation by Profile</a>
+      [[DX-PROF-CONNEG]] describes a mechanism that separates content-type
+      negotiation from requesting a response conforming to a specific information model,
+      whether that model is a specific standard, specification or profile.
+      The <a data-cite="DX-PROF-CONNEG#requestsandresonses"
+             >operations of the abstract model</a> 
+      are "list profiles" and "get resource by profile".
+      The "get resource" operation can provide the profile requests with HTTP
+      header <code>Accept-Profile</code> and responses use <code>Content-Profile</code>, or
+      alternatively the request can use HTTP Query String parameters.
```
TallTed, line 2690: How does this address the clear statement (I don't have the citation handy; sorry) that profiled media types are supposed to be subsets of the media type, i.e., `<< >>` notation is not valid `text/turtle` regardless of the profile attached -- _because `<< >>` notation **is not valid** `text/turtle`!_

cg-spec/editors_draft.html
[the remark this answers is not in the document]
gkellogg: According to [RFC6838](https://datatracker.ietf.org/doc/html/rfc6838#page-20), a provisional registration may be appropriate.

> Accordingly, a provisional registration process is provided to
   support early assignment of media type names in the standards tree.
...
> Provisional registrations MAY be updated or abandoned at any time.
   When the registration is abandoned, the media type is no longer
   registered in any sense; it can subsequently be registered just like
   any other unassigned media type name.

But, there's the theory vs. practice issue of ever encouraging the use of a provisional media type.

cg-spec/editors_draft.html
pchampin: @TallTed  I agree. However, the way I see it, a future working group could
- extend text/turtle to support turtle-star 
- define different profiles for the text/turtle (with or without RDF-star, possibly others)
- define that the default profile SHOULD be "without RDF-star". Since old client will not require any profile, this would encourage servers to accommodate them if practical.

As Andy wrote, this does not solve the optimistic/pessimistic dilemma for us. But this should be considered by a future WG in the design of a durable solution.

cg-spec/editors_draft.html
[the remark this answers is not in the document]
pchampin: @gkellogg 
I would definitely not take the risk of registering a media-type that was provisionally registered by someone else and then abandoned, even if the RFC allows me to do so! This feels like seeking trouble...

@TallTed 
"temporary" can be understood as a form a "local", in the temporal dimension... ;-)

cg-spec/editors_draft.html
afs: The negotiation is of information model and "profile" in that document is a "specification that constrains, extends, combines, or provides guidance or explanation about the usage of other specifications."

That said, TR/dx-prof-conneg is of possible if `text/turtle` is updated. Media types are not immutable and other types have changed "in place" over the years or bypassed the media type issues and done their own thing (e.g. XML 1.1).

As we have identified, there is no perfect solution here. Each route has negative impact in some cases. Until we have field experience, we can't make a fact-based decision.

TallTed: > That said, `TR/dx-prof-conneg` is [only] possible if `text/turtle` is updated. Media types are not immutable and other types have changed "in place" over the years or bypassed the media type issues and done their own thing (e.g. XML 1.1).
> 
> As we have identified, there is no perfect solution here. Each route has negative impact in some cases. Until we have field experience, we can't make a fact-based decision.

Fair enough. But since we can't make a fact-based decision, I submit that we should not make a decision at all, but rather list the possible options we know of, along with the pros and cons and additional requirements of each — such that it's clear that, for instance, `TR/dx-prof-conneg` could be brought into play (and solve our media type quandary) *IFF* `text/turtle` (and some other current media types) were also updated to have a profile which resulted in today's format and which was the effective profile when no profile was declared, *along with* such other profiles as were deemed needed/relevant at that point, such as (hopefully) profiles for RDF-star serializations, such as (the most relevant for this example) Turtle-star.

afs: I don't believe it makes a decision.

The section is "non-normative".  

The new text says "This may be useful when the information is available in RDF-star or in non-RDF-star RDF 1.1 form," and refers back to the dilemma set out in the beginning of the section.

TallTed: A section being non-normative does not mean it will not be take as a strong suggestion, which can amount to a decision even though authors and editors and others loudly assert that it was not a decision.

I think that links back to these discussions/issues/PRs would be as appropriate, possibly more so, as the links to RFCs we're not deciding to use but suggesting as possibilities...

While minimizing the possibility of confusion, or suggestion of debate, or even word count, can sometimes be appropriate in WG Candidate or Proposed Recommendations, I do not think the same applies to CG Reports, especially where we (at least, I) anticipate significant discussion and conclusions that may or may not include any of the possibilities listed in the report, due to the anticipated (or at least hoped for) broader community participation.

pchampin:
This was discussed during yesterday's call: https://w3c.github.io/rdf-star/Minutes/2021-06-04.html#t06

PR #179, opened 2021-06-04
Move nt-ttl syntax test to the Turtle area.

afs:
This PR moves the tests that are running N-Triples syntax tests as Turtle tests into the turtle/syntax area.

This leaves tests for an NT parser in `nt/syntax` and an NT-only parser system does not need to react to the file extension or test type.

This PR branched off PR #178.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/pull/179.html" title="Last updated on Jun 4, 2021, 4:29 PM UTC (9d2112e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/179/5b20a3f...9d2112e.html" title="Last updated on Jun 4, 2021, 4:29 PM UTC (9d2112e)">Diff</a>


afs:
Check failure in EARL report generation because the PR moves tests and so old earl reports are invalidated.

This is a cost of this PR.

(Other tests have been added since the last round of EARL report submissions.)

gkellogg:
> Check failure in EARL report generation because the PR moves tests and so old earl reports are invalidated.
> 
> This is a cost of this PR.
> 
> (Other tests have been added since the last round of EARL report submissions.)

This will fail until the individual reports are updated. We could delete the ones which reference the old test, as they'll need to be re-submitted anyway, or just ignore the errors.

gkellogg (approved):

afs:
@gkellogg - do you think the test are better in the "turtle" area (as per PR) because they are `.ttl` or "ntriples" area (original) because they are about ntriples.

I'm neutral.

This came up from @pchampin's n-triples only (currently) implementation experience of picking up the n-triples tests only and getting confused.

gkellogg:
> @gkellogg - do you think the test are better in the "turtle" area (as per PR) because they are `.ttl` or "ntriples" area (original) because they are about ntriples.

Well, N-Triples is, of course, a subset of Turtle, and it's entirely appropriate for Turtle tests (.ttl) to contain files which are restricted to that subset. Repeating N-Triples tests as Turtle tests can pose a maintenance burden, but simplifies instructions for the implementor. JSON-LD did quite a but of test duplication, particularly for Expansion and ToRdf tests, and it's worked out okay.

As an implementor, I use a single implementation for running N-Triples, Turtle and TriG tests (rdf-trig), so as to make a single EARL submission, but it could just as well be done each from a different implementation (e.g., the code-path for the native N-Triples parser is different than the Turtle/TriG parser), but that doesn't seem to relevant to the purpose of the implementation report.

afs:
@gkellogg - Thanks for updating the HTML manifests.

Issue #197, opened 2021-07-18
Add more metadata in the tests/manifest.ttl


gatemezing:
I would love to see more metadata in the [manifest.ttl](tests/manifest.ttl)  file. The minimum ones would be the date of creation (`dct:issued`), last modification (`dct:modified`), version if possible and the license (`dct:license`) info attached with a property.

afs:
License is in the usual place: top of the repo : https://github.com/w3c/rdf-star/blob/main/LICENSE.md - also in the top of the manifest.ttl as a comment.

Whatever else, please don't remove these because this is the usual practice for open source software.

gatemezing:
Sure @afs ! I'll propose something as a PR with the comments on top of the file

afs:
Great! There will be details to discuss.

gatemezing:
Perfect!! It's a good starting point to reach a consensus :)

TallTed:
Very important to codefence all `@`things here on github, else people like `@prefix` get pinged with updates about which they do not care.

I also recommend switching from `@prefix x: <http://example.com/ex/> .` notation to `PREFIX x: <http://example.com/ex/>`, partly but not only because it helps avoid that `@`handle problem.

pchampin:
I guess this can be closed, this #217 was merged?

PR #198, opened 2021-07-18
suggest metadata in manifest turtle file

gatemezing:
This a based on the suggestion of the issue #197.

### DIFF
--- a/tests/manifest.ttl
+++ b/tests/manifest.ttl
@@ -1,17 +1,27 @@
-## Distributed under both the "W3C Test Suite License" [1]
-## and the "W3C 3-clause BSD License".
 ## [1] https://www.w3.org/Consortium/Legal/2008/04-testsuite-license
 ## [2] https://www.w3.org/Consortium/Legal/2008/03-bsd-license
 
-PREFIX rdf:    <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
-PREFIX rdfs:   <http://www.w3.org/2000/01/rdf-schema#>
-PREFIX mf:     <http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#>
-PREFIX rdft:   <http://www.w3.org/ns/rdftest#>
-PREFIX trs:    <https://w3c.github.io/rdf-star/tests#>
+@PREFIX rdf:    <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@PREFIX rdfs:   <http://www.w3.org/2000/01/rdf-schema#> .
+@PREFIX mf:     <http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#> .
+@PREFIX rdft:   <http://www.w3.org/ns/rdftest#> .
+@PREFIX trs:    <https://w3c.github.io/rdf-star/tests#> .
+@PREFIX dct: <http://purl.org/dc/terms/> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+@prefix foaf: <http://xmlns.com/foaf/0.1/> .
 
 trs:manifest  rdf:type mf:Manifest ;
-    rdfs:label "RDF-star test suite" ;
-    rdfs:seeAlso <https://w3c.github.io/rdf-tests/> ;
+  rdfs:label "RDF-star test suite"@en ;
+	rdfs:label "La suite des tests pour RDF-star"@fr;
+	rdfs:label "Conjuto de pruebas para RDF-star"@es;
+	dct:issued "2021-06-21"^^xsd:date ; #not sure if this was the first release date
+	dct:creator [ foaf:name "Gregg Kellog" ] ; #maybe add foaf:homepage
+	dct:creator [ foaf:name "Andy Seaborne" ];
+  rdfs:seeAlso <https://w3c.github.io/rdf-tests/> ;
+	dct:modified "2021-07-18"^^xsd:date ; # change to the exact date
+	dct:licence <https://www.w3.org/Consortium/Legal/2008/04-testsuite-license>;
+	dct:licence <https://www.w3.org/Consortium/Legal/2008/03-bsd-license> ;
+	dct:contributor [ foaf:homepage <https://w3c.github.io/rdf-star/> ; foaf:name " W3C RDF-star Working Group" ] ;
     mf:include (
       <nt/syntax/manifest.ttl>
       <semantics/manifest.ttl>


tests/manifest.ttl
```diff
@@ -1,17 +1,27 @@
...
+  rdfs:label "RDF-star test suite"@en ;
+	rdfs:label "La suite des tests pour RDF-star"@fr;
+	rdfs:label "Conjuto de pruebas para RDF-star"@es;
+	dct:issued "2021-06-21"^^xsd:date ; #not sure if this was the first release date
+	dct:creator [ foaf:name "Gregg Kellog" ] ; #maybe add foaf:homepage
+	dct:creator [ foaf:name "Andy Seaborne" ];
+  rdfs:seeAlso <https://w3c.github.io/rdf-tests/> ;
+	dct:modified "2021-07-18"^^xsd:date ; # change to the exact date
+	dct:licence <https://www.w3.org/Consortium/Legal/2008/04-testsuite-license>;
+	dct:licence <https://www.w3.org/Consortium/Legal/2008/03-bsd-license> ;
```
afs, line 23: :-)

RDF is conjunctive - these two `dct:licence` that say _both_ licenses apply, not a choice of one of them!

The root LICENSE is "Contributions to Test Suites are made under the W3C 3-clause BSD License" so drop the testsuite-license as it does not apply.

Background:
The testsuite-license is not an open source license (there is no unrestricted right to modify granted). The idea being test suite for working group is what is produced, not a downstream modified form there of. Hence the bsd-license, which is open source.

tests/manifest.ttl
```diff
@@ -1,17 +1,27 @@
...
-    rdfs:label "RDF-star test suite" ;
-    rdfs:seeAlso <https://w3c.github.io/rdf-tests/> ;
+  rdfs:label "RDF-star test suite"@en ;
+	rdfs:label "La suite des tests pour RDF-star"@fr;
+	rdfs:label "Conjuto de pruebas para RDF-star"@es;
+	dct:issued "2021-06-21"^^xsd:date ; #not sure if this was the first release date
+	dct:creator [ foaf:name "Gregg Kellog" ] ; #maybe add foaf:homepage
+	dct:creator [ foaf:name "Andy Seaborne" ];
+  rdfs:seeAlso <https://w3c.github.io/rdf-tests/> ;
+	dct:modified "2021-07-18"^^xsd:date ; # change to the exact date
```
afs, line 21: For the community to discuss: 

- dates get out of date (they are more maintenance work)
- the same date may not apply to all tests.

https://www.w3.org/ns/rdftest# has some per-test vocabulary.

tests/manifest.ttl
```diff
@@ -1,17 +1,27 @@
...
+@prefix foaf: <http://xmlns.com/foaf/0.1/> .
 
 trs:manifest  rdf:type mf:Manifest ;
-    rdfs:label "RDF-star test suite" ;
-    rdfs:seeAlso <https://w3c.github.io/rdf-tests/> ;
+  rdfs:label "RDF-star test suite"@en ;
+	rdfs:label "La suite des tests pour RDF-star"@fr;
+	rdfs:label "Conjuto de pruebas para RDF-star"@es;
+	dct:issued "2021-06-21"^^xsd:date ; #not sure if this was the first release date
+	dct:creator [ foaf:name "Gregg Kellog" ] ; #maybe add foaf:homepage
```
afs, line 18: I don't think we should assert who "created" the suite - the important point is the community owns and maintains the work (much like the trend not to have "author" in open source code).

tests/manifest.ttl
```diff
@@ -1,17 +1,27 @@
...
-## and the "W3C 3-clause BSD License".
 ## [1] https://www.w3.org/Consortium/Legal/2008/04-testsuite-license
 ## [2] https://www.w3.org/Consortium/Legal/2008/03-bsd-license
 
-PREFIX rdf:    <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
-PREFIX rdfs:   <http://www.w3.org/2000/01/rdf-schema#>
-PREFIX mf:     <http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#>
-PREFIX rdft:   <http://www.w3.org/ns/rdftest#>
-PREFIX trs:    <https://w3c.github.io/rdf-star/tests#>
+@PREFIX rdf:    <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
```
afs, line 4: `PREFIX`, no '.'.

i.e. RDF 1.1 style Turtle.

tests/manifest.ttl
```diff
@@ -1,17 +1,27 @@
...
+@PREFIX rdfs:   <http://www.w3.org/2000/01/rdf-schema#> .
+@PREFIX mf:     <http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#> .
```
gkellogg, line 6: It's either `PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>` or `@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .` Although some processors may be more lenient. The SPARQL form does not have a training `.`.

tests/manifest.ttl
gkellogg: In any case, I'm not the sole creator. It could go on each test, but that would be quite a burden. Date can be useful, if it is properly maintained, but it should be replicated to the JSON-LD and HTML versions of the manifest, as well (along with other bits) to keep the different versions of the test manifests isomorphic.

tests/manifest.ttl
gkellogg: Per-test metadata will never be managed properly. It's more useful to maintain the modification date of the manifest, itself, which could be useful to ensure that EARL reports match the latest version of the test manifest. IIRC, @pchampin suggested something like this elsewhere.

gkellogg (changes requested):

tests/manifest.ttl
afs: As long as modified !=> created in any way (another reason to remove dct:creator -- there isn't a neat "create" point for a manifest).

The https://www.w3.org/ns/rdftest#  vocabulary has been used for per test when the WG formally accepted tests. I agree that in the community space, maintenance is burden.

tests/manifest.ttl
gatemezing: ACK! good to know. I'll change that.

tests/manifest.ttl
gatemezing: > In any case, I'm not the sole creator. It could go on each test, but that would be quite a burden. Date can be useful, if it is properly maintained, but it should be replicated to the JSON-LD and HTML versions of the manifest, as well (along with other bits) to keep the different versions of the test manifests isomorphic.

Re creator, I can just put the community group. I agree that the date should be replicated in all the files. We are not having many modifications for this set of manifest files. I can't imagine more than 3 modifications/updates in the lifetime of a WG. I'm in favour of properly publish such assets with proper metadata.

gatemezing: > I don't think we should assert who "created" the suite - the important point is the community owns and maintains the work (much like the trend not to have "author" in open source code).

Well, the same can be applied to the HTML document. We have editors, authors and contributors. Why don't we also add them in the RDF files we are publishing?

gatemezing (commented):
I understand the suggestions. I will wait for a general discussion during a group meeting after this summer's break.

tests/manifest.ttl
TallTed: Per-test metadata *should* be manageable, and I think setting that example may be a subtle goodness we can bring to the general community.  `dct:creator`, `dct:editor`, `dct:modified`, `dct:created` are all reasonable.  I have certainly had times where such info would have been valuable to me, about RDF artifacts as well as HTML, TXT, PDF, etc.

tests/manifest.ttl
TallTed: I also prefer the `PREFIX …` to the `@prefix … .` notation.

Further, I suggest ordering the list by the prefix (most "alpha sort lines" tools will do the right thing) -- because it helps avoid duplicate declarations, and makes human review of the file easier.

tests/manifest.ttl
gkellogg: It's easy enough to support additional properties on a test entry, but the code that transforms to JSON-LD and HTML will require some minor updates to make sure this doesn't get lost. Presumably, much metadata could be discovered through GitHub commit history, and automating this might be the best way to go, at least to get existing tests seeded with data.

gkellogg:
CI fails due to syntax errors in some of the manifest files. Basically, you need to be able to run tests/make-jsonld-manifests.rb and tests/make-heml-manifests.py successfully, but it should work with valid Turtle inputs. There may be more required to properly reflect added properties in JSON-LD and HTML views.

tests/semantics/manifest.ttl
```diff
@@ -1,12 +1,30 @@
...
-PREFIX mf:     <http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#>
-PREFIX test:   <http://www.w3.org/2001/sw/DataAccess/tests/test-dawg#>
-PREFIX xsd:    <http://www.w3.org/2001/XMLSchema#>
-PREFIX trs:    <https://w3c.github.io/rdf-star/tests/semantics#>
+## Distributed under both the "W3C Test Suite License" [1]
+## and the "W3C 3-clause BSD License".
+## [1] https://www.w3.org/Consortium/Legal/2008/04-testsuite-license
+## [2] https://www.w3.org/Consortium/Legal/2008/03-bsd-license
+
+@PREFIX rdf:    <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
```
afs, line 6: I prefer `PREFIX`.

Is there a reason for `@` style?

But if we must use `@`, the `@prefix`.

tests/semantics/manifest.ttl
```diff
@@ -1,12 +1,30 @@
...
+@PREFIX dct: <http://purl.org/dc/terms/> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+@prefix foaf: <http://xmlns.com/foaf/0.1/> .
+
 
 trs:manifest a mf:Manifest;
-  rdfs:label "RDF-star Semantics tests";
+  rdfs:label "RDF-star Semantics tests"@en ;
+  rdfs:label "La suite des tests pour la sémantique de RDF-star"@fr;
+	rdfs:label "Conjunto de pruebas para la semántica de RDF-star"@es;
```
afs, line 19: Alignment error.

Please use spaces not tabs.

Here, your tabs are not the same a GH tabs.

tests/trig/eval/manifest.ttl
```diff
@@ -11,7 +11,16 @@ PREFIX rdft:   <http://www.w3.org/ns/rdftest#>
 PREFIX trs:    <https://w3c.github.io/rdf-star/tests/trig/eval#>
 
 trs:manifest  rdf:type mf:Manifest ;
-    rdfs:label "TriG-star Evaluation Tests" ;
+    rdfs:label "TriG-star Evaluation Tests"@en ;
+    rdfs:label "La suite des tests pour TriG-star"@fr;
+	 rdfs:label "Conjunto de pruebas para Trig-star"@es;
```
afs, line 16: Double alignment error.

tests/trig/eval/manifest.ttl
```diff
@@ -11,7 +11,16 @@ PREFIX rdft:   <http://www.w3.org/ns/rdftest#>
...
+	dct:issued "2021-06-21"^^xsd:date ; #not sure if this was the first release date
+	dct:creator [ foaf:name "Gregg Kellog" ] ; #maybe add foaf:homepage
+	dct:creator [ foaf:name "Andy Seaborne" ];
```
afs, line 19: I don't think we should name people. The output is from the community. creator compared to contributor are not so easily distinguished.

tests/turtle/syntax/manifest.ttl
```diff
@@ -3,15 +3,24 @@
...
-PREFIX mf:     <http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#>
-PREFIX test:   <http://www.w3.org/2001/sw/DataAccess/tests/>
-PREFIX rdft:   <http://www.w3.org/ns/rdftest#>
-PREFIX trs:    <https://w3c.github.io/rdf-star/tests/turtle/syntax#>
-
+PREFIX rdf:    <http://www.w3.org/1999/02/22-rdf-syntax-ns#> 
+PREFIX rdfs:   <http://www.w3.org/2000/01/rdf-schema#> 
+PREFIX mf:     <http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#> 
+PREFIX rdft:   <http://www.w3.org/ns/rdftest#> 
+PREFIX trs:    <https://w3c.github.io/rdf-star/tests#>
```
afs, line 10: Alignment.

afs (changes requested):
My comments on one file apply equally to the other manifests.

tests/semantics/manifest.ttl
gatemezing: Yes, I fix that locally. I'll just use `PREFIX`

tests/trig/eval/manifest.ttl
gatemezing: OK. I'll just mention the name of the group using `dct:creator` and remove the `dct:contributor`

tests/semantics/manifest.ttl
gatemezing: I'm not sure to understand, but I'll find how to solve this. Thanks.

tests/turtle/syntax/manifest.ttl
TallTed: @afs -- What alignment are you looking for?  This appears to me to have a reasonable alignment, though it's not exactly to my taste.  I think you'll be more likely to get your desired alignment if you specify it.

My own preference is to alpha-sort the prefixes (helps avoid accidental duplication), then left-align `PREFIX`, right-align the prefixes on their `:`, and left align the URLs.

afs: Whitespace.
In the PR, the tabs cause problems in GH interface and in some places there is extra spaces as well.

pchampin:
This was discussed during today's call: https://w3c.github.io/rdf-star/Minutes/2021-09-03.html#t03

afs:
I took a diff from the current state of this PR and applied it to a local up-to-date repo.

The changes for sparq/, trig/, turtle/ *.ttl parse and the tests run.

1. The semantics/manifest did not merge - possibly the PR is based off an older version?
2. There don't seem to be changes to `sparql/eval`.
3. Alignment issues.
4. Development comments.
5. Additional trailing whitespace has been added

_semantic/manifest.ttl_

"git apply semantics.diff" gives:
```
semantics.diff:17: trailing whitespace.
PREFIX rdf:    <http://www.w3.org/1999/02/22-rdf-syntax-ns#> 
semantics.diff:18: trailing whitespace.
PREFIX rdfs:   <http://www.w3.org/2000/01/rdf-schema#> 
semantics.diff:19: trailing whitespace.
PREFIX mf:     <http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#> 
semantics.diff:20: trailing whitespace.
PREFIX rdft:   <http://www.w3.org/ns/rdftest#> 
semantics.diff:21: trailing whitespace.
PREFIX trs:    <https://w3c.github.io/rdf-star/tests#> 
error: patch failed: tests/semantics/manifest.ttl:1
error: tests/semantics/manifest.ttl: patch does not apply
```

_Alignment_: the problem show when the files are viewed including in this PR.

Tabs width are an unknown (2? 4? 8? or "smart tabs" in an IDE) - it is better to use tabs/spaces consistently and the files have spaces.

On linux, the `expand` command will do most of the work.

This is what I see in sparql-syntax/manifest.ttl with tab width of 8:
```
trs:manifest  rdf:type mf:Manifest ;
    rdfs:label "SPARQL-star Syntax Tests"@en ;
    rdfs:label "La suite des tests pour SPARQL-star"@fr;
        rdfs:label "Conjunto de pruebas para SPARQL-star"@es;
        dct:issued "2021-06-21"^^xsd:date ; #not sure if this was the first release date
    rdfs:seeAlso <https://w3c.github.io/rdf-tests/> ;
        dct:modified "2021-07-18"^^xsd:date ; # change to the exact date
        dct:licence <https://www.w3.org/Consortium/Legal/2008/03-bsd-license> ;
        dct:creator [ foaf:homepage <https://w3c.github.io/rdf-star/> ; foaf:name " W3C RDF-star Working Group" ] ;
    mf:entries
```
In trig/eval/manifest (which for some reason has an indent of 3 and was probably supposed to be 4 - ditto turtle/eval):

```
trs:manifest  rdf:type mf:Manifest ;
   rdfs:label "TriG-star Evaluation Tests"@en ;
   rdfs:label "La suite des tests pour TriG-star"@fr;
   rdfs:label "Conjunto de pruebas para Trig-star"@es;
    dct:issued "2021-06-21"^^xsd:date ; #not sure if this was the first release date
   rdfs:seeAlso <https://w3c.github.io/rdf-tests/> ;
    dct:modified "2021-07-18"^^xsd:date ; # change to the exact date
    dct:licence <https://www.w3.org/Consortium/Legal/2008/03-bsd-license> ;
    dct:creator [ foaf:homepage <https://w3c.github.io/rdf-star/> ; foaf:name " W3C RDF-star Working Group" ] ;
    mf:entries
```

_Development comments_

These appear in several files:  
` #not sure if this was the first release date`  
` # change to the exact date`

These shouldn't in the merge.

_Trailing whitespace_

An example taken from from tests/semantics/manifest.ttl but it seems to be in all files: when PREFIX is added there a trailing space on URIs:
```
+PREFIX rdf:    <http://www.w3.org/1999/02/22-rdf-syntax-ns#>_
```
where _ is an added space. This happens on added and existing prefixes so it ials changes to lines not otherwise changed.

gatemezing:
> I took a diff from the current state of this PR and applied it to a local up-to-date repo.
> 
Thanks for your time Andy! 
> The changes for sparq/, trig/, turtle/ *.ttl parse and the tests run.
> 
> 1. The semantics/manifest did not merge - possibly the PR is based off an older version?
> 2. There don't seem to be changes to `sparql/eval`.
> 3. Alignment issues.
> 4. Development comments.
> 5. Additional trailing whitespace has been added
> 
I'll try to address the points in my next commits. Thanks.

tests/trig/eval/manifest.ttl
gatemezing: Pushed a new version

gatemezing:
@afs, any hint on this conflict issue with the turtle/eval/manifest file? TIA

afs:
Whitespace difference.

TallTed:
@gatemezing -- fwiw, I often find it's easier to do the resolve-conflicts thing by copying the conflicting section into an external editor which can display "invisible characters" like tabs and spaces (I use BBEdit on macOS), splitting it into two local docs, and running local diffs to reconcile them (which sometimes means changes flow both ways).  When they're identical, I copy and paste one back to Github.

gatemezing:
> @gatemezing -- fwiw, I often find it's easier to do the resolve-conflicts thing by copying the conflicting section into an external editor which can display "invisible characters" like tabs and spaces (I use BBEdit on macOS), splitting it into two local docs, and running local diffs to reconcile them (which sometimes means changes flow both ways). When they're identical, I copy and paste one back to Github.

Thanks @TallTed ! I'm struggling in Windows with Notepad++.

gatemezing:
> Whitespace difference.

Thanks for the hint! Next, it's to make ruby works to validate the isomorphism file. Some hours of testing ahead :)

gatemezing:
> CI fails due to syntax errors in some of the manifest files. Basically, you need to be able to run tests/make-jsonld-manifests.rb and tests/make-heml-manifests.py successfully, but it should work with valid Turtle inputs. There may be more required to properly reflect added properties in JSON-LD and HTML views.

@gkellogg - Yeah, I'm trying to make the tests using make-jsonld-manifests.rb. Is there a way to use ruby to convert the .ttl file to jsonld? 
BTW, I just realized that when validating the tests/manifest-context.jsonld using [online playground](https://json-ld.org/playground/) I get an error `jsonld.SyntaxError: Invalid JSON-LD syntax; "@id" value must a string.` Is that normal?

gkellogg:
I’ll take a look at it and suggest changes. It’s likely that the json-led frame is not handling multi-language literals (which will also affect the html generation).

gatemezing:
> I’ll take a look at it and suggest changes. It’s likely that the json-led frame is not handling multi-language literals (which will also affect the html generation).

Thanks! I made some changes in the manifest-context.jsold file to take into account those multi-lingual labels. I can't just figure out from the errors output, where exactly to make the chances. If you find the fix for one file, I can take care to make rest of the changes for the remaining files.

gkellogg:
The reason for the failure of the semantics manifest is that the "test" prefix was removed. This would be the only manifest that still uses it (for test:approval). It happens that the JSON-LD framing depends on this being there to be able to show manifests that are not included in `mf:entries`. This also creates a parsing error, as the "test" prefix is used without being defined. So, the simplest thing would be to re-introduce the prefix definition:

```
PREFIX test:   <http://www.w3.org/2001/sw/DataAccess/tests/test-dawg#>
```

Alternatively:

* semantics/manifest.ttl uses `test:approval` without defining the `test` prefix. It's the only manifest that uses this, so I suggest just removing from the semantics manifest. This would also indicate removing the "approval" and "statusOf" properties from manifest-context.jsonld. (May be some implications for test:withdrawn entries).
* Comment out the following unused tests:
  * trs:malformed-literal-bnode
  * trs:malformed-literal-bnode-2
  * trs:malformed-literal-bnode-3
  * trs:triple-in-object
  * trs:triple-in-subject

Do that, and make-jsonld-manifests.rb completes satisfactorily. The HTML manifests also build, but I didn't check them to be sure they display all the pertinent information.

gatemezing:
> So, the simplest thing would be to re-introduce the prefix definition:

Hi @gkellogg, this solved the errors. I've not checked the html output. Thanks again!

tests/semantics/manifest.ttl
gatemezing: solved

afs:
This needs some tidy up.

1. It looks like it is based off the repo as of about 9 weeks ago.
2. There are a lot of intermediate commits (about 17).
3. There are conflicts on `tests/turtle/eval/manifest.ttl`, `tests/trig/eval/manifest.ttl`.

Either

A. Squash the PR to one commit first, pull current main to main in your local repo and then rebase on the PR - only one set of conflicts to handle otherwise there will be conflict after conflict as each individual commit are merged.

Or

B. It might be better to take a diff of the changes, start with an up-to-date fork of the main branch, apply the diff (fixing any problems) and making a new PR or force pushing onto this PR.

gatemezing:
> This needs some tidy up.
> 
> 1. It looks like it is based off the repo as of about 9 weeks ago.
> 2. There are a lot of intermediate commits (about 17).
> 3. There are conflicts on `tests/turtle/eval/manifest.ttl`, `tests/trig/eval/manifest.ttl`.
> 
> Either
> 
> A. Squash the PR to one commit first, pull current main to main in your local repo and then rebase on the PR - only one set of conflicts to handle otherwise there will be conflict after conflict as each individual commit are merged.
> 
> Or
> 
> B. It might be better to take a diff of the changes, start with an up-to-date fork of the main branch, apply the diff (fixing any problems) and making a new PR or force pushing onto this PR.

Thanks for the hints! I started with option A and sent yet-another-PR. I did not see any conflicts with the main branch. Maybe I'm missing something here. Ufff, quite challenging to make a normal PR in this repo :)

afs:
Running the manifest.ttl files, I get suite names in a mix of English, French and Spanish. The code looks for one rdfs:label.

Suggestion:
Use rdfs:label with non-lang-tag text.
Use skos for the language tagged literals.

gatemezing:
> Running the manifest.ttl files, I get suite names in a mix of English, French and Spanish. The code looks for one rdfs:label.
> 
> Suggestion:
> Use rdfs:label with non-lang-tag text.
> Use skos for the language tagged literals.

Thanks for the suggestions. However, I'm afraid to enter in a new loop with the changes you suggested, that is update both .ttl and .jsonld files, etc. Could the editors safely do that once the PR is merged?

Issue #169, opened 2021-05-14
Occurrences vocabulary


rat10:
After long debate it was decided that embedded triples refer to an ("abstract") triple, not to any specific occurrence. Use cases like the seminal provenance use case [0] however need to refer to a specific occurrence to document e.g. when a certain triple was inserted into a certain graph. The draft report so far contains an example [1] that shows how an occurrence can be derived from a triple, using a made up property in the notorious `example.org` namespace.

The question now is: do we properly define this property as part of the RDF-star vocabulary?

**Pro**: this is an important use case and we should provide more than just an non-committal example.
**Con**: the property alone is not enough. A complete solution would also have to provide a means to describe the graph in which the triple occurs. Otherwise the provided solution is just as underspecified as RDF standard reification. This however is relatively uncharted territory and touches e.g. the mined area of named graphs.

[0] https://w3c.github.io/rdf-star/cg-spec/2021-04-13.html#the-seminal-example
[1] https://w3c.github.io/rdf-star/cg-spec/2021-04-13.html#occurrences-example

rat10:
Maybe "uncharted territory" is too pessimistic. It is well defined that an RDF graph is a set of RDF triples. Even if there is no term in the RDF vocabulary that denotes graphs we could provide two properties:
```
occurrenceOf
inGraph
```
and advise that the range of inGraph is an IRI pointing to a set of triples e.g. in a document or a named graph. We are just addressing a set of triples, so no need to get into discussions about what that set means or entails or about what a dataset that eventually contains it means etc etc. What could possibly go wrong?!

hartig:
I am in favor of adding such a vocabulary.

However, @rat10, you end your description of this issue with the question: "What could possibly go wrong?!" Now, this makes me wonder what the purpose of adding this question is. Is this meant to be a rhetorical question? Do you forsee anything that might "go wrong" if we define such a vocabulary?

rat10:
@hartig The sentence "What could possibly go wrong?!" is rethorical, meant to express that "I'm not sure I have thought this through sufficiently. I don't see any problems but this area is notorious for non-obvious problems."

afs:
As this is likely a common need:  `<< :a :b :c >> rdfx:occursIn :SomeGraphIRI .`

rat10:
@afs Renaming the property `rdfx:inGraph` to `rdfx:occurrsIn` we get
```
_:x rdfx:occurrenceOf << :a :b :c >>;
[...]

PR #204, opened 2021-09-04
Classes for RDF-star terminology

afs:
Some initial material taken from #169 from @pchampin.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/pull/204.html" title="Last updated on Sep 8, 2021, 12:57 PM UTC (3dc0281)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/204/7dd0ab6...3dc0281.html" title="Last updated on Sep 8, 2021, 12:57 PM UTC (3dc0281)">Diff</a>

### DIFF
--- a/cg-spec/editors_draft.html
+++ b/cg-spec/editors_draft.html
@@ -3059,6 +3059,32 @@ style="color:black;font-weight:bold">sameTerm</code></dfn></b>
       This issue proposes to introduce a vocabulary for locally overriding the <a href="#ref-opacity">opaque</a> semantics of <a>quoted triples</a>, making them transparent in some contexts.
     </div>
 
+    <div class="defn">
+      <pre>## Vocabulary for concepts used by or introduced by RDF-star.
+PREFIX rdf:       &lt;http://www.w3.org/1999/02/22-rdf-syntax-ns#&gt;
+PREFIX rdfs:      &lt;http://www.w3.org/2000/01/rdf-schema#&gt;
+PREFIX rdf-star:  &lt;http://www.w3.org/ns/rdf-star#&gt;
+
+rdf-star:Triple      a rdfs:Class ;
+    rdfs:isDefinedBy   &lt;http://www.w3.org/ns/rdf-star#&gt;
+    rdfs:label         "RDF Triple" ;
+    rdfs:comment       "The class of RDF triples." ;
+    rdfs:subClassOf    rdfs:Resource .
+
+rdf-star:Graph       a rdfs:Class ;
+    rdfs:isDefinedBy   &lt;http://www.w3.org/ns/rdf-star#&gt;
+    rdfs:label         "RDF Graph" ;
+    rdfs:comment       "The class of RDF graphs." ;
+    rdfs:subClassOf    rdfs:Resource .
+
+rdf-star:Source      a rdfs:Class ;
+    rdfs:isDefinedBy   &lt;http://www.w3.org/ns/rdf-star#&gt;
+    rdfs:label         "RDF Source" ;
+    rdfs:seeAlso       &lt;https://www.w3.org/TR/rdf11-concepts/#h3_change-over-time&gt; ;
+    rdfs:comment       "The class of RDF Sources." ;
+    rdfs:subClassOf    rdfs:Resource .
+      </pre>
+    </div>
   </section>
 
   <section class="appendix informative">
--- a/cg-spec/editors_draft.html
+++ b/cg-spec/editors_draft.html
@@ -3067,19 +3067,19 @@ PREFIX rdf-star:  &lt;http://www.w3.org/ns/rdf-star#&gt;
 
 rdf-star:Triple      a rdfs:Class ;
     rdfs:isDefinedBy   &lt;http://www.w3.org/ns/rdf-star#&gt;
-    rdfs:label         "RDF Triple" ;
-    rdfs:comment       "The class of RDF triples." ;
+    rdfs:label         "RDF-star Triple" ;
+    rdfs:comment       "The class of RDF-star triples." ;
     rdfs:subClassOf    rdfs:Resource .
 
 rdf-star:Graph       a rdfs:Class ;
     rdfs:isDefinedBy   &lt;http://www.w3.org/ns/rdf-star#&gt;
-    rdfs:label         "RDF Graph" ;
-    rdfs:comment       "The class of RDF graphs." ;
+    rdfs:label         "RDF-star Graph" ;
+    rdfs:comment       "The class of RDF-star graphs." ;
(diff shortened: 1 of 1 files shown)


hartig:
Good! A few comments/questions:

Given that these IRIs are in an RDF-star related "namespace", shouldn't `rdf-star:Triple` be the class of all RDF-star triples rather than RDF triples? Similarly for `rdf-star:Graph`.

Regarding `rdf-star:Source` I am not sure. Is "RDF source" a concept that is actually defined anywhere?

afs:
This comes from material in https://github.com/w3c/rdf-star/issues/169#issuecomment-842078440 with a tiny data and HTML bit of formatting. With these large issues, discussions can skip over the small (and uncontroversial?!) parts.

RDF Source is from [RDF 1.1 Concepts](https://www.w3.org/TR/rdf11-concepts/#h3_change-over-time) as informal terminology but becomes significant in discussion #169. There is a similar situation is SPARQL Update where a graph store is a collection of slots and the contents of the slot can be changed.

Yes - it is RDF-star triples and RDF-star graphs. It is in the namespace "rdf-star" so I think we can use a shorter form ":Graph" as a convenience but we should "RDF-star triples" in the rdfs:label and rdfs:comment which I've now changed.

hartig:
Thanks for the changes in the labels and comments. That's what I was after ;-)

Thanks also for the reference regarding RDF source. I wasn't aware of that one.

cg-spec/editors_draft.html
```diff
@@ -3059,6 +3059,32 @@ <h2>RDF-star vocabulary</h2>
...
+    <div class="defn">
+      <pre>## Vocabulary for concepts used by or introduced by RDF-star.
+PREFIX rdf:       &lt;http://www.w3.org/1999/02/22-rdf-syntax-ns#&gt;
+PREFIX rdfs:      &lt;http://www.w3.org/2000/01/rdf-schema#&gt;
+PREFIX rdf-star:  &lt;http://www.w3.org/ns/rdf-star#&gt;
+
+rdf-star:Triple      a rdfs:Class ;
+    rdfs:isDefinedBy   &lt;http://www.w3.org/ns/rdf-star#&gt;
+    rdfs:label         "RDF-star Triple" ;
+    rdfs:comment       "The class of RDF-star triples." ;
```
gkellogg, line 3071: Is this the class of quoted triples, triples containing a quoted triple, or the superset of all RDF triples which may or may not contain an embedded triple? If the latter, should there be a class for quoted triples?

afs: It is the class of triples in RDF-star. A triple in [RDF 1.1 concepts](https://www.w3.org/TR/rdf11-concepts/#section-triples) is three components. The three components define the triple. It can be in more than one graph. Asserted => element of a graph set ; in RDFstar, quoted => may also in an RDF term.

I wondered about a class of quoted triples and came to conclusion that a triple can be quoted or asserted but it exists regardless.

Also, if a quoted triple is different then there is a 1-1 correspondence between quoted and assertable triples. i.e. it makes little difference.

hartig: I had asked myself the same question and came to the same conclusion as @afs. There is no need to add a class for quoted triples. An RDF-star triple is unique identified by its three components, no matter where it appears.

cg-spec/editors_draft.html
```diff
@@ -3059,6 +3059,32 @@ <h2>RDF-star vocabulary</h2>
...
+    rdfs:isDefinedBy   &lt;http://www.w3.org/ns/rdf-star#&gt;
+    rdfs:label         "RDF-star Graph" ;
+    rdfs:comment       "The class of RDF-star graphs." ;
+    rdfs:subClassOf    rdfs:Resource .
+
+rdf-star:Source      a rdfs:Class ;
+    rdfs:isDefinedBy   &lt;http://www.w3.org/ns/rdf-star#&gt;
+    rdfs:label         "RDF-star Source" ;
+    rdfs:seeAlso       &lt;https://www.w3.org/TR/rdf11-concepts/#h3_change-over-time&gt; ;
+    rdfs:comment       "The class of RDF Sources." ;
```
TallTed, line 3084:
```suggestion
    rdfs:comment       "The class of RDF-star sources." ;
```

cg-spec/editors_draft.html
rat10: I tried parsing @afs answer to @gkellogg but I didn't succeed because of the mixture of apparent typos, unfinished sentences and probably also some ignorance on my side. Especially this here I'm unable to parse: `in RDFstar, quoted => may also in an RDF term` , but also this: `Also, if a quoted triple is different then there is a 1-1 correspondence between quoted and assertable triples. i.e. it makes little difference`. Could you please again explain?

afs: "if a quoted triple is different (from an asserted triple) then there is an isomorphism between quoted and assertable triples"

rat10:
You skipped some of the terms proposed by Pierre-Antoine in [#169](https://github.com/w3c/rdf-star/issues/169#issuecomment-842078440) and you completely ignored my comments. Would you please explain the rationale for the terms you defined, what their use case is? Also the explaining comments are so terse that they seem quite useless to me.

cg-spec/editors_draft.html
rat10: On my book all quoted triples are assertable but not all assertable triples are quoted because a quoted triple requires the act of quoting a (assertable, not necessarily asserted) triple. Quoting a triple is an act that creates something that in RDF-star syntax looks like << s p o >>. This we call quoted triple, formerly we called it embedded triple. I think this needs a class and it is distinct from all triples that are assertable.

rat10: Or maybe you mean not quoted, but "quotable", the class of all triples that can be quoted? That is indeed isomorphic to the class of all assertable triples. Which brings me back to my question below: what made you choose these terms and not others? What are the use cases?

afs: > all quoted triples are assertable
Surely it's "all triples are assertable"? One does not asserted a quoted triple. I think this is what your second comment says? 

There are triples.

They can be used as quoted (RDF terms) or asserted (elements of a set of triples making up a graph) or both.

["RDF-star introduces quoted triple, which is a new kind of RDF term."](https://w3c.github.io/rdf-star/cg-spec/editors_draft.html#quoted-triples).

pchampin:
This was discussed during today's call: https://w3c.github.io/rdf-star/Minutes/2021-09-17.html#t03
Resolved to merge.

cg-spec/editors_draft.html
rat10: Ok, now I understand.

PR #213, opened 2021-10-13
Improve intro figure

pchampin:
New text in the boxes, which hopefully makes their relationship more obvious.

### DIFF
--- a/cg-spec/editors_draft.html
+++ b/cg-spec/editors_draft.html
@@ -375,35 +375,38 @@ INSERT {
           <tr>
             <td rowspan=3>
               <div class="box">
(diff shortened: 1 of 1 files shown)


afs:
Nice.

If using text to create the linkages, then having the section two box at the top with five boxes below:

```
                           § 2. Concepts and Abstract Syntax
                                can be...

... queried with                      ... interpreted through         ... represented by
§ 4. SPARQL-star Query Language       § 6. RDF-star Semantics         § 7. RDF-star vocabulary

... updated with                       ... serialized with
§ 5. SPARQL-star Update                § 3. Concrete Syntaxes
```

PR #216, opened 2021-10-15
Fix the rdf:type of a couple of the Turtle syntax tests

kal:
Tests 'nt-ttl-star-4' and 'nt-ttl-star-5' should both be typed as 'rdft:TestTurtlePositiveSyntax' rather than 'rdft:TestNTriplesPositiveSyntax' in the Turtle syntax test suite.

### DIFF
--- a/tests/turtle/syntax/manifest.jsonld
+++ b/tests/turtle/syntax/manifest.jsonld
@@ -225,13 +225,13 @@
     },
     {
       "@id": "trs:nt-ttl-star-4",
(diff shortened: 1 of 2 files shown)


afs:
@pchampin -- this is a small bug fix. It has no copyright or patent implications. Do we need the IPR test to pass? Can I mark it "non-substantive"?

@kal - explanation - we are operating as part of the RDF-DEV CG. The contributions checking process is slanted towards content for the documents where people submit under the W3C IP declaration for CGs.

afs:
@kal - thanks!

Issue #219, opened 2021-10-29
referential opacity


pfps:
There has been a lot of discussion in the CG on referential transparency / opacity and decisions made in this area that affect the behaviour of RDF*.

If a WG is going to go forward there should be re-examination of this issue.  However, the draft report pushes this important concept into the technical details of the semantics.

There should be discussion of the effects of the current semantics of RDF* in the introductory material of the document.   This will help WG members understand what RDF* is and, probably more importantly, is not.

pchampin:
this was discussed during today's meeting: https://w3c.github.io/rdf-star/Minutes/2021-11-12.html#t04

pchampin:
Closing as the discussion is continued on #225.

PR #225, opened 2021-12-02
Add a section in intro about "Common misconceptions"

pchampin:
This addresses issues #64 and #219.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/pull/225.html" title="Last updated on Dec 6, 2021, 5:42 PM UTC (9640ada)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/225/44ba3f4...9640ada.html" title="Last updated on Dec 6, 2021, 5:42 PM UTC (9640ada)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/pull/225.html" title="Last updated on Dec 15, 2021, 8:32 AM UTC (46f6a46)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-star/225/a2d93d8...46f6a46.html" title="Last updated on Dec 15, 2021, 8:32 AM UTC (46f6a46)">Diff</a>

### DIFF
--- a/cg-spec/editors_draft.html
+++ b/cg-spec/editors_draft.html
@@ -301,7 +301,7 @@
     <section class="informative">
       <h2>Overview</h2>
 
-<p>This specification describes RDF-star, an extension of RDF's conceptual data model and concrete syntaxes, providing a more compact form of reification. This model and syntaxes enable the creation of concise triples that reference other triples as subject and object resources.
+<p>This specification describes RDF-star, an extension of RDF's conceptual data model and concrete syntaxes, providing a compact alternative to <a href="https://www.w3.org/TR/rdf11-mt/#reification">standard reification</a>. This model and syntaxes enable the creation of concise triples that reference other triples as subject and object resources.
 Triples that include a triple as a subject or an object are known as <a>RDF-star triples</a>. The following dataset shows the example RDF-star triples from above using the <a href="#turtle-star">Turtle-star</a> syntax, which uses double angle brackets to enclose a triple serving as a subject or object resource:</p>
 
       <pre data-transform="updateExample"
@@ -318,7 +318,7 @@ PREFIX :    <http://www.example.org/>
 
 <p>After declaring a prefix so that IRIs can be abbreviated, the first triple in this example asserts that employee38 has a familyName of "Smith". Note that this dataset does not assert that employee38 has a jobTitle of "Assistant Designer"; it says that employee22 has made that claim. In other words, the triple "employee38 has a jobTitle of 'Assistant Designer'" is not what we call an <a>asserted triple</a>, like "employee38 has a familyName of 'Smith'" above; rather, it is known as a <a>quoted triple</a>.</p>
 
-<p>If we added the triple `:employee38 :jobTitle "Assistant Designer"` below the triple about employee22's claim in the example above, then this triple about employee38's jobTitle would be both a <a>quoted</a> and an <a>asserted</a> triple.
+<p>If we added the triple `:employee38 :jobTitle "Assistant Designer"` in the example above, then this triple about employee38's jobTitle would be both a <a>quoted</a> and an <a>asserted</a> triple.
 This pattern is quite common, so Turtle-star offers a dedicated syntax for it, called the <a>annotation syntax</a>, illustrated in <a href="#example-annotation"></a> below. Note that this construct is purely syntactic sugar, as it can be expanded using only the double angle brackets.</p>
 
 <pre data-transform="updateExample"
@@ -352,7 +352,7 @@ This pattern is quite common, so Turtle-star offers a dedicated syntax for it, c
 PREFIX : <http://www.example.org/> 
 
 SELECT ?claimer WHERE {
-   ?claimer :claims << :employee38 ?property ?value >>
+   << :employee38 ?property ?value >> :accordingTo ?claimer
 }
 -->
 </pre>
--- a/cg-spec/editors_draft.html
+++ b/cg-spec/editors_draft.html
@@ -675,7 +675,7 @@ INSERT {
         <!--
         :lizTaylor :married :richardBurton {| :occurrence
             [ :year 1964; :location :montreal ],
-            [ :year 1975, :location :botswana ]
+            [ :year 1975; :location :botswana ]
         |}.
         -->
       </pre>
--- a/cg-spec/editors_draft.html
+++ b/cg-spec/editors_draft.html
@@ -375,6 +375,21 @@ INSERT {
 -->
 </pre>
 
+    </section>
+    <section>
+      <h2>Common misconceptions</h2>
+
+      <p>Many people come to RDF-star with wrong ideas about what it is or how to use it. This section lists and corrects the most frequent ones.
+
+      <p>RDF-star is <em>not</em> syntactic sugar for <a href="https://www.w3.org/TR/rdf11-mt/#reification">standard reification</a>. While the latter is a vocabulary fitting in the standard RDF model (abstract syntax), RDF-star extends that model (see <a href="#concepts"></a>) with a new construct, namely <a>quoted triples</a>. Note also that the two are not mutually exclusive: standard reification can still be used in conjunction with RDF-star.
+
+      <p>Unlike reified statements, RDF-star <a>quoted triples</a> are unique: wherever `&lt;&lt; :employee38 :jobTitle "Assistant Designer" &gt;&gt;` appears, it denotes one and the same thing. This impacts the modelling choices one has to make when using RDF-star; this is further discussed in <a href="#occurrences"></a>.
+
+      <p>Finally, <a>quoted triples</a> are referentially opaque. This means that two <a>quoted triples</a> containing distinct but equivalent terms (i.e. known to denote the same thing) are nonetheless considered to be different triples. <a href="#ref-opacity"></a> and the following sections discuss the rationale for this design, and how some form of referential transparency can still be added to RDF-star if and when needed.
+    </section>
+
+    <section>
+      <h2>Conventions</h2>
 <p>For the remainder of this document, examples will assume that the following prefixes have been declared to represent the IRIs shown with them here: </p>
 
 <table>
--- a/cg-spec/editors_draft.html
+++ b/cg-spec/editors_draft.html
@@ -379,7 +379,7 @@ INSERT {
     <section>
       <h2>Common misconceptions</h2>
 
-      <p>Many people come to RDF-star with wrong ideas about what it is or how to use it. This section lists and corrects the most frequent ones.
+      <p>People coming to RDF-star may have wrong ideas about what it is or how to use it. The purpose of this section is to dispel the most common misconceptions.
 
       <p>RDF-star is <em>not</em> syntactic sugar for <a href="https://www.w3.org/TR/rdf11-mt/#reification">standard reification</a>. While the latter is a vocabulary fitting in the standard RDF model (abstract syntax), RDF-star extends that model (see <a href="#concepts"></a>) with a new construct, namely <a>quoted triples</a>. Note also that the two are not mutually exclusive: standard reification can still be used in conjunction with RDF-star.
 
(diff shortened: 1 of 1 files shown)


cg-spec/editors_draft.html
```diff
@@ -375,6 +375,21 @@ <h2>Overview</h2>
 -->
 </pre>
 
+    </section>
+    <section>
+      <h2>Common misconceptions</h2>
+
+      <p>Many people come to RDF-star with wrong ideas about what it is or how to use it. This section lists and corrects the most frequent ones.
+
+      <p>RDF-star is <em>not</em> syntactic sugar for <a href="https://www.w3.org/TR/rdf11-mt/#reification">standard reification</a>. While the latter is a vocabulary fitting in the standard RDF model (abstract syntax), RDF-star extends that model (see <a href="#concepts"></a>) with a new construct, namely <a>quoted triples</a>. Note also that the two are not mutually exclusive: standard reification can still be used in conjunction with RDF-star.
```
ericprud, line 384: > standard reification can still be used in conjunction with RDF-star

If anyone's got some compelling examples of this, it could be nice to add a section. (someday, not in this PR)

pchampin: I guess my point was to say: "if you are using standard reification, you don't need to change anything to migrate to RDF-star, your reifications will still "work" as before."

Are there concrete use-cases for using std reification together with quoted triples, beyond supporting legacy RDF? I am not entirely sure, because RDF-star offers a more compact way to express the same thing. With the following property:
```
:reificationOf a rdf:Property, rdf-star:TransparencyEnablingProperty ;
    rdfs:domain rdf:Statement ;
    rdfs:range rdf-star:Triple.
```
One could replace
```
[] a rdf:Statement ;
   rdf:subject _:s ;
   rdf:predicate :p ;
   rdf:object "o".
```
with
```
[] :reificationOf << _:s :p "o" >>.
```

cg-spec/editors_draft.html
```diff
@@ -626,6 +641,20 @@ <h2>Triples and occurrences</h2>
...
+        class="nohighlight example"
+      >
+        <!--
+        :lizTaylor :married :richardBurton {| :occurrence
+            [ :year 1964; :location :montreal ],
+            [ :year 1975, :location :botswana ]
+        |}.
+        -->
+      </pre>
+
```
hartig: I am not sure whether including this example here is a good idea. It can be misleading for two reasons: First, the `:occurrence` property used here has a completely different meaning than the inverse of the `:occurrenceOf` property used in the previous example; in particular, the `:occurrence` property here is not about the occurrence of triples (which is the topic of this section) but about the occasion of a marriage event. Second, this example assumes referential transparency (i.e., the `:occurrence` property should be a TEP).

cg-spec/editors_draft.html
```diff
@@ -375,6 +375,21 @@ <h2>Overview</h2>
 -->
 </pre>
 
+    </section>
+    <section>
+      <h2>Common misconceptions</h2>
+
+      <p>Many people come to RDF-star with wrong ideas about what it is or how to use it. This section lists and corrects the most frequent ones.
```
hartig, line 382:
```suggestion
      <p>People coming to RDF-star may have wrong ideas about what it is or how to use it. The purpose of this section is to dispel the most common misconceptions.
```

cg-spec/editors_draft.html
pchampin: I agree that the similarity in names may be misleading (this was also raised by Thomas on the mailing list), and should be fixed. Maybe `':event` would be a better choice of name ?

I would like to keep the example, though, precisely because it demonstrates that the type-token distinction shows up in different use-cases:
- unasserted or asserted triples
- with "opaque occurrences" or with "transparent occurrences" (because yes, `:occurrence`/`:event` in this example would be a TEP)

pchampin:
This was discussed during today's call: https://w3c.github.io/rdf-star/Minutes/2021-12-03.html#x106

cg-spec/editors_draft.html
```diff
@@ -318,7 +318,7 @@ <h2>Overview</h2>
 
 <p>After declaring a prefix so that IRIs can be abbreviated, the first triple in this example asserts that employee38 has a familyName of "Smith". Note that this dataset does not assert that employee38 has a jobTitle of "Assistant Designer"; it says that employee22 has made that claim. In other words, the triple "employee38 has a jobTitle of 'Assistant Designer'" is not what we call an <a>asserted triple</a>, like "employee38 has a familyName of 'Smith'" above; rather, it is known as a <a>quoted triple</a>.</p>
 
-<p>If we added the triple `:employee38 :jobTitle "Assistant Designer"` below the triple about employee22's claim in the example above, then this triple about employee38's jobTitle would be both a <a>quoted</a> and an <a>asserted</a> triple.
+<p>If we added the triple `:employee38 :jobTitle "Assistant Designer"` in the example above, then this triple about employee38's jobTitle would be both a <a>quoted</a> and an <a>asserted</a> triple.
```
TallTed, line 321:
```suggestion
<p>If we added the triple `:employee38 :jobTitle "Assistant Designer"` to the example above, then this triple about employee38's jobTitle would be both a <a>quoted</a> and an <a>asserted</a> triple.
```

cg-spec/editors_draft.html
```diff
@@ -375,6 +375,21 @@ <h2>Overview</h2>
 -->
 </pre>
 
+    </section>
+    <section>
+      <h2>Common misconceptions</h2>
+
+      <p>People coming to RDF-star may have wrong ideas about what it is or how to use it. The purpose of this section is to dispel the most common misconceptions.
+
+      <p>RDF-star is <em>not</em> syntactic sugar for <a href="https://www.w3.org/TR/rdf11-mt/#reification">standard reification</a>. While the latter is a vocabulary fitting in the standard RDF model (abstract syntax), RDF-star extends that model (see <a href="#concepts"></a>) with a new construct, namely <a>quoted triples</a>. Note also that the two are not mutually exclusive: standard reification can still be used in conjunction with RDF-star.
```
TallTed, line 384:
```suggestion
      <p>RDF-star is <em>not</em> syntactic sugar for <a href="https://www.w3.org/TR/rdf11-mt/#reification">standard reification</a>. While the latter is a vocabulary that fits into the standard RDF model (abstract syntax), RDF-star extends that model (see <a href="#concepts"></a>) with a new construct, namely <a>quoted triples</a>. Note also that the two are not mutually exclusive: standard reification can be used in conjunction with RDF-star.
```

cg-spec/editors_draft.html
```diff
@@ -375,6 +375,21 @@ <h2>Overview</h2>
...
+
+      <p>Unlike reified statements, RDF-star <a>quoted triples</a> are unique: wherever `&lt;&lt; :employee38 :jobTitle "Assistant Designer" &gt;&gt;` appears, it denotes one and the same thing. This impacts the modelling choices one has to make when using RDF-star; this is further discussed in <a href="#occurrences"></a>.
```
TallTed, line 386:
```suggestion
      <p>Unlike reified statements, but like URIs, RDF-star <a>quoted triples</a> are unique: wherever `&lt;&lt; :employee38 :jobTitle "Assistant Designer" &gt;&gt;` appears, it always denotes one and the same thing. This impacts the modeling choices one has to make when using RDF-star; this is further discussed in <a href="#occurrences"></a>.
```

cg-spec/editors_draft.html
```diff
@@ -375,6 +375,21 @@ <h2>Overview</h2>
...
+
+      <p>Finally, <a>quoted triples</a> are referentially opaque. This means that two <a>quoted triples</a> containing distinct but equivalent terms (i.e. known to denote the same thing) are nonetheless considered to be different triples. <a href="#ref-opacity"></a> and the following sections discuss the rationale for this design, and how some form of referential transparency can still be added to RDF-star if and when needed.
```
TallTed, line 388:
```suggestion
      <p>Finally, again like URIs, <a>quoted triples</a> are referentially opaque. This means that two <a>quoted triples</a> containing distinct but equivalent terms (i.e., coreferences, known to denote the same thing) are nonetheless considered to be different triples. <a href="#ref-opacity"></a> and the following sections discuss the rationale for this design, and how some form of referential transparency can still be added to RDF-star if and when needed.
```

cg-spec/editors_draft.html
pchampin:
```suggestion
      <p>Unlike reified statements, RDF-star <a>quoted triples</a> are unique: wherever `&lt;&lt; :employee38 :jobTitle "Assistant Designer" &gt;&gt;` appears, it always denotes one and the same thing. This impacts the modeling choices one has to make when using RDF-star; this is further discussed in <a href="#occurrences"></a>.
```

pchampin: @TallTed I didn't keep the "but like URIs" from your suggestion above. First, because it should be "IRIs" :wink:. Second, because the reference to IRIs rather than, say, literals, might be confusing. Finally, because the comparison of quoted triples with IRIs (or literals, for that matter) becomes slippery when quoted triples contain blank nodes (although of course, that does not change anything to their "unicity" described here).

cg-spec/editors_draft.html
pchampin: I think the "again like URIs/IRIs" is incorrect. IRIs in RDF (and in RDF-star when used outside quoted triples) are referentially transparent. They can be replaced by any synonymous term without changing the meaning of the graph.

TallTed: > [@pchampin] I think the "again like URIs/IRIs" is incorrect. IRIs in RDF (and in RDF-star when used outside quoted triples) are referentially transparent. They can be replaced by any synonymous term without changing the meaning of the graph.

Yeah, sorry, "lexically opaque" is what I was meaning, not "referentially opaque". My brain regularly flips these two. I imagine others' brains do the same, which is one reason I usually argue against using "referentially opaque" as a descriptor. Revising suggested change.

cg-spec/editors_draft.html
```diff
@@ -301,7 +301,7 @@ <h2>Background and Motivation</h2>
     <section class="informative">
       <h2>Overview</h2>
 
-<p>This specification describes RDF-star, an extension of RDF's conceptual data model and concrete syntaxes, providing a more compact form of reification. This model and syntaxes enable the creation of concise triples that reference other triples as subject and object resources.
+<p>This specification describes RDF-star, an extension of RDF's conceptual data model and concrete syntaxes, providing a compact alternative to <a href="https://www.w3.org/TR/rdf11-mt/#reification">standard reification</a>. This model and syntaxes enable the creation of concise triples that reference other triples as subject and object resources.
```
TallTed, line 304: @pchampin -- Sorry I didn't catch this before merge... I can make a PR, if you prefer.
```suggestion
<p>This specification describes RDF-star, an extension of RDF's conceptual data model and concrete syntaxes, which provides a compact alternative to <a href="https://www.w3.org/TR/rdf11-mt/#reification">standard RDF reification</a>. This model and its syntaxes enable the creation of concise triples that reference other triples as subject and object resources.
```

pchampin: No problem. I can apply it directly on `master`.
Question though: I don't really like "this model and *its* syntaxes" because the idea is to refer to everything that was mentioned in the previous sentence, i.e. the model AND the syntaxes. I understand that the singular "this" is a problem, but would "these model and syntaxes" work? Or "this model and these syntaxes"?

TallTed: I'd be OK with "this model and these syntaxes", though to me it implies that no other syntaxes can "enable ... concise ... resources."

To my mind, "This model" and _any syntax that expresses it_ (which includes but is not limited to those we're currently specifying) can "enable ... concise ... resources." hence, "its syntaxes".

pchampin: I get your point. "its syntax" it is, then :)
Done in 7ff7179

Issue #233, opened 2021-12-18
Test report generator failure


afs:
In an up-to-date repo, `bundle install` and `bundle exec rake` run fine.

But I tried to add a new EARL report `apache-jena.ttl` I get errors relating to `"http://champin.net/#pa"` which isn't in the file.
(draft [apache-jena.ttl.txt](https://github.com/w3c/rdf-star/files/7739236/apache-jena.ttl.txt) attached - with `.txt` added for GH).

How can I debug this?

_Update_: PR #235 works - the difference is "homepage" URIs changed from "http:" to "https:" and if the timestamp of the file is later than all files in the directory. Restoring earl.* from git does not leave timestamps in the right state.

```
Traceback (most recent call last):
	9: from /home/afs/gems/bin/earl-report:23:in `<main>'
	8: from /home/afs/gems/bin/earl-report:23:in `load'
	7: from /var/lib/gems/2.7.0/gems/earl-report-0.7.0/bin/earl-report:100:in `<top (required)>'
	6: from /var/lib/gems/2.7.0/gems/earl-report-0.7.0/lib/earl_report.rb:479:in `generate'
	5: from /var/lib/gems/2.7.0/gems/earl-report-0.7.0/lib/earl_report.rb:559:in `earl_turtle'
	4: from /var/lib/gems/2.7.0/gems/earl-report-0.7.0/lib/earl_report.rb:559:in `each'
	3: from /var/lib/gems/2.7.0/gems/earl-report-0.7.0/lib/earl_report.rb:565:in `block in earl_turtle'
	2: from /var/lib/gems/2.7.0/gems/earl-report-0.7.0/lib/earl_report.rb:565:in `each'
	1: from /var/lib/gems/2.7.0/gems/earl-report-0.7.0/lib/earl_report.rb:566:in `block (2 levels) in earl_turtle'
/var/lib/gems/2.7.0/gems/earl-report-0.7.0/lib/earl_report.rb:574:in `ttl_entity': undefined method `map' for "http://champin.net/#pa":String (NoMethodError)
```

If I rename the file to `jena.ttl, I get 
```
Traceback (most recent call last):
	11: from /home/afs/gems/bin/earl-report:23:in `<main>'
	10: from /home/afs/gems/bin/earl-report:23:in `load'
	 9: from /var/lib/gems/2.7.0/gems/earl-report-0.7.0/bin/earl-report:100:in `<top (required)>'
[...]

PR #234, opened 2021-12-18
Add report: apache-jena.ttl

afs:
This just adds the EARL TTL file for Apache Jena.

Regenerating locally ran into #233.

---

Update: no luck - https://github.com/w3c/rdf-star/runs/4569494409?check_suite_focus=true


afs:
Bad PR.

PR #251, opened 2022-02-10
Add missing objects in `sparql-star-syntax-bad-ann-path-*.rq` tests

rodentrabies:
Object parts are missing in the patterns in tests

    sparql-star-syntax-bad-ann-path-2.rq
    sparql-star-syntax-bad-ann-path-3.rq
    sparql-star-syntax-bad-ann-path-4.rq
    sparql-star-syntax-bad-ann-path-5.rq
    sparql-star-syntax-bad-ann-path-6.rq

This makes a correct SPARQL-star parser fail the tests as intended,
but for a different reason.

### DIFF
--- a/tests/sparql/syntax/sparql-star-syntax-bad-ann-path-2.rq
+++ b/tests/sparql/syntax/sparql-star-syntax-bad-ann-path-2.rq
@@ -1,5 +1,5 @@
 PREFIX : <http://example.com/ns#>
 
 SELECT * {
-  :x (:p|:q) {| ?p ?o |}
+  :x (:p|:q) :o {| ?p ?o |}
 }
--- a/tests/sparql/syntax/sparql-star-syntax-bad-ann-path-3.rq
+++ b/tests/sparql/syntax/sparql-star-syntax-bad-ann-path-3.rq
@@ -1,5 +1,5 @@
 PREFIX : <http://example.com/ns#>
 
 SELECT * {
-  :x :p* {| ?p ?o |}
+  :x :p* :o {| ?p ?o |}
 }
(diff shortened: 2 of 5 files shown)


gkellogg:
These test files are intentionally bad examples and a parser should raise an error, as indicated by the type of the test entry in the manifest.

rodentrabies:
Isn't the intention of these tests to show that it doesn't make sense to have annotation syntax if the verb in the pattern is a property path (same as in `sparql-star-syntax-bad-ann-path-1.rq`)?

gkellogg:
The grammar allows annotations of patterns using property paths, so that in itself would parse okay. But @afs wrote the tests originally so is in the best position to comment on the intention of the tests.

pchampin:
> The grammar allows annotations of patterns using property paths, so that in itself would parse okay.

I think that @rodentrabies' point is precisely that: these test files do *not* respect the grammar, because they miss an object after the property path.

hartig:
> The grammar allows annotations of patterns using property paths, so that in itself would parse okay.

While the grammar allows such annotations, they are still invalid. We have a restriction related to these cases, which is in the text below the grammar in Section 4.2 of the spec. 

https://www.w3.org/2021/12/rdf-star.html#sparql-star-grammar

afs (approved):
@rodentrabies Thank you!

rodentrabies:
No problem! Do I need to do anything else for this to get merged? I received email from iprbot about W3C account, but I think this change goes into "Typos" category.

gkellogg (approved):

pchampin:
this was discussed during today's call: https://w3c.github.io/rdf-star/Minutes/2022-02-11.html#x131

PR #259, opened 2022-04-27
adding morph-kgc implementation

dachafra:
If this implementation isn't suitable for the webpage, please let us know where we can contribute with it :-)
Also closing `tr` and `td` tags in the table.

PR #260, opened 2022-04-28
adding morph-kgc implementation

dachafra:
only adding the implementation, as was requested in #259


afs:
This is a change to the implementation doc. 

@pchampin Do we need IPR?

PR #266, opened 2022-06-16
 Improve definition of function sparql-compare.

### LINKED ISSUE #263: sparql-compare type signature conflicts with description
The type signature for [sparql-compare](https://www.w3.org/2021/12/rdf-star.html#sparql-compare) is:

```
xsd:boolean   sparql-compare (RDF-star term, RDF-star term)
```

with a return value of `xsd:boolean`. However, the description says:

> The function sparql-compare returns -1, 0, 1, or throws an error, based on the comparison of two RDF-star terms.

I believe `xsd:boolean` should be changed to `xsd:integer` here.

### LINKED ISSUE #264: Confusing definition of sparql-compare for triples with IRI components
The first step of [sparql-compare](https://www.w3.org/2021/12/rdf-star.html#sparql-compare) is defined as:

> If neither A nor B is an RDF-star triple term, compare by SPARQL 1.1 operators <, =, >) and return the comparison value (-1, 0, +1) or throw an error as defined by SPARQL 1.1.

I find this confusing in cases where the operator is `=` and the operands are IRIs. Looking at the [sparql-star-op-3](https://w3c.github.io/rdf-star/tests/sparql/eval/#sparql-star-op-3) test for the `:x7` test record, we have:

`<< :a :b 9 >>` `<` ` << :a :b 123 >>`

In this case, `<` is not defined over IRI operands, and so would result in an error if invoked. The test case seems to assume that `sparql-compare` is able to compare subject, predicate, and object values, and determine that the overall expression is true because the object value comparison `9 < 123` is true. I imagine this indicates that `sparql-compare`'s first rule needs to *first* check for equality using `=` and only invoke `<` or `>` if the operands are *not* equal.

If this interpretation is correct, I think the wording of the `sparql-compare` definition needs to be explicit about avoiding `<` and `>` in comparison of IRI values which are equal.

### LINKED ISSUE #265: Possible issue with sparql-compare semantics and ORDER BY
[SPARQL 15.1](https://www.w3.org/TR/sparql11-query/#modOrderBy) says:

> The "<" operator (see the Operator Mapping and 17.3.1 Operator Extensibility) defines the relative order of pairs of numerics, simple literals, xsd:strings, xsd:booleans and xsd:dateTimes. Pairs of IRIs are ordered by comparing them as simple literals.

This suggests that the `<` operator does not handle IRIs, even within ORDER BY handling. The ordering of IRIs is handled as a special case in evaluating `OrderBy`.

But [RDF-star 4.4.11 Triple terms with ORDER BY](https://www.w3.org/2021/12/rdf-star.html#triple-term-comparison) says:

> RDF-star triple terms are compared using the `<` operator.

[RDF-star 4.4.10](https://www.w3.org/2021/12/rdf-star.html#rdf-star-operator-mapping) adds an entry to the operator mappings for RDF triple terms:

> `A < B`	`RDF triple term`	`RDF triple term`	`op:numeric-equal(sparql-compare(A, B), -1)`	`xsd:boolean`

and [RDF-star 4.4.8 sparql-compare](https://www.w3.org/2021/12/rdf-star.html#sparql-compare) is defined recursively with:

* non-triple terms are compared "by SPARQL 1.1 operators `<`, `=`, `>`"
* "If `sparql-compare(SUBJECT(A), SUBJECT(B)) != 0`, then return this value" (and similarly for predicate and object)

I think these things combined suggest that when evaluating an `ORDER BY`, IRIs that are contained within RDF-star triple terms are not comparable (but that they are intended to be).

I believe the [sparql-star-order-2](https://w3c.github.io/rdf-star/tests/sparql/eval/#sparql-star-order-2) test relies on the intended semantics here, but fails as written.

I think the definitions need to be updated to allow the special IRI comparability introduced in SPARQL 15.1 to apply to the non-triple term rule in sparql-compare *when evaluating within the context of an OrderBy operator*.

afs:
This closes #263.
This closes #264.
This closes #265.

### DIFF
--- a/cg-spec/editors_draft.html
+++ b/cg-spec/editors_draft.html
@@ -2287,8 +2287,9 @@ style="color:black;font-weight:bold">sameTerm</code></dfn></b>
       <section id="sparql-compare">
         <h3>sparql-compare</h3>
         
-        <p>The function <code>sparql-compare</code> is defined to support value-comparison
-          operators such as <code>=</code> and <code>ORDER BY</code>.
+        <p>
+          The function <code>sparql-compare</code> is defined to support value-comparison
+          operators, such as <code>=</code> and <code>&lt;</code>.
         </p>
         <p>
           <code>xsd:integer</code> &nbsp;
@@ -2301,8 +2302,9 @@ style="color:black;font-weight:bold">sameTerm</code></dfn></b>
           error, based on the comparison of two
           <a href="#dfn-rdf-star-terms">RDF-star terms</a>.
           When comparing two RDF-star triple terms, comparison is 
-          by pairwise comparison of subject, then predicate, then object.
-        </p>
+          by pairwise comparison of components subject, then predicate, then object.
+          An error is thrown if comparing two components throws an error.
+        </p> 
         <div class="defn">
           <p>
             <b>Definition: 
@@ -2311,19 +2313,34 @@ style="color:black;font-weight:bold">sameTerm</code></dfn></b>
             </b>
           </p>
           <ul>
-            <li>If neither A nor B is an RDF-star triple term,
-              compare by SPARQL 1.1 operators
-              <code>&lt;</code>, <code>=</code>, <code>&gt;</code>)
-              and return the comparison value (`-1`, `0`, `+1`) or throw an error
-              as defined by SPARQL 1.1.<br/>
+            <li>
+              If neither A nor B is an RDF-star triple term, apply the
+              <a href="https://www.w3.org/TR/sparql11-query/#OperatorMapping"
+                 >the SPARQL 1.1 operator mappings</a>.
+              If none apply, raise an error.
+              Return `-1`, `0`, or `+1` based on which
+              of <code>&lt;</code>, <code>=</code> and <code>&gt;</code>
+              returns true.
             </li>
-            <li>If either A or B is an RDF-star triple term, and the other is not 
-              an RDF-star triple term, then error.</li>
-            <li>If <code>sparql-compare(SUBJECT(A), SUBJECT(B)) != 0</code>,
-              then return this value.</li>
-            <li>If <code>sparql-compare(PREDICATE(A), PREDICATE(B)) != 0</code>,
-              then return this value.</li>
-            <li>Return <code>sparql-compare(OBJECT(A), OBJECT(B))</code></li>
+            <li>
+              If either A or B is an RDF-star triple term, and the other is not 
+              an RDF-star triple term, then raise an error.
+            </li>
+            <li>
+              Evaluate <code>sparql-compare(SUBJECT(A), SUBJECT(B))</code>.
+              If the evalution causes an error, raise an error.
+              If the evaluation result is -1 or +1, return this value.
+            </li>
+            <li>
+              Evaluate <code>sparql-compare(PREDICATE(A), PREDICATE(B)).</code>
+              If the evalution causes an error, raise an error.
+              If the evaluation result is -1 or +1, return this value.
+            </li>
+            <li>
+              Evaluate <code>sparql-compare(OBJECT(A), OBJECT(B))</code>.
+              If the evalution causes an error, raise an error.
+              otherwise return the evaluation value.
+            </li>
           </ul>
         </div>
       </section>
--- a/cg-spec/editors_draft.html
+++ b/cg-spec/editors_draft.html
@@ -2389,7 +2389,6 @@ style="color:black;font-weight:bold">sameTerm</code></dfn></b>
           and covers cases involving literals.
         </p>
       </section>
-
       <section id="rdf-star-operator-mapping">
         <h3>Operator Mappings</h3>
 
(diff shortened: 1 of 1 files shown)


gkellogg (commented):
Doesn’t seem to be loading the EBNF files, where the path has changed. Otherwise, looks good.

gkellogg:
> Doesn’t seem to be loading the EBNF files, where the path has changed. Otherwise, looks good.

Never mind, that looks like a diff issue, preview seems okay.

gkellogg (approved):

afs:
The PR version has the same 6 warnings 5 errors as the main version. 

Load `<div data-include="../ebnf/n-quads.html"></div>` when viewed from checked out git repo gets "NetworkError when attempting to fetch resource" -- maybe because it's a file: URL.

cg-spec/editors_draft.html
```diff
@@ -2301,8 +2302,9 @@ <h3>sparql-compare</h3>
           error, based on the comparison of two
           <a href="#dfn-rdf-star-terms">RDF-star terms</a>.
           When comparing two RDF-star triple terms, comparison is 
-          by pairwise comparison of subject, then predicate, then object.
-        </p>
+          by pairwise comparison of components subject, then predicate, then object.
```
hartig, line 2305:
```suggestion
          by pairwise comparison of their components, first their subject, next their predicate, finally their object.
```

cg-spec/editors_draft.html
```diff
@@ -2301,8 +2302,9 @@ <h3>sparql-compare</h3>
...
+          An error is thrown if comparing two components throws an error.
```
hartig, line 2306:
```suggestion
          An error is thrown if any of these three comparisons throws an error.
```

cg-spec/editors_draft.html
```diff
@@ -2311,19 +2313,34 @@ <h3>sparql-compare</h3>
...
+            </li>
+            <li>
+              Evaluate <code>sparql-compare(SUBJECT(A), SUBJECT(B))</code>.
+              If the evalution causes an error, raise an error.
+              If the evaluation result is -1 or +1, return this value.
+            </li>
+            <li>
+              Evaluate <code>sparql-compare(PREDICATE(A), PREDICATE(B)).</code>
+              If the evalution causes an error, raise an error.
+              If the evaluation result is -1 or +1, return this value.
```
hartig, line 2337:
```suggestion
              If the evaluation result is `-1` or `+1`, return this value.
            </li>
            <li>
              Evaluate <code>sparql-compare(PREDICATE(A), PREDICATE(B)).</code>
              If the evalution causes an error, raise an error.
              If the evaluation result is `-1` or `+1`, return this value.
```

cg-spec/editors_draft.html
```diff
@@ -2311,19 +2313,34 @@ <h3>sparql-compare</h3>
...
+            </li>
+            <li>
+              Evaluate <code>sparql-compare(OBJECT(A), OBJECT(B))</code>.
+              If the evalution causes an error, raise an error.
+              otherwise return the evaluation value.
```
hartig, line 2342:
```suggestion
              otherwise return the evaluation result.
```

hartig (approved):
Just a few minor editorial cosmetics.

cg-spec/editors_draft.html
TallTed, line 2305:
```suggestion
          by pairwise comparison of their components: first of their subjects; then of their predicates; then of their objects.
```

cg-spec/editors_draft.html
TallTed, line 2342:
```suggestion
              Otherwise, return the evaluation result.
```

cg-spec/editors_draft.html
```diff
@@ -2477,9 +2493,10 @@ <h3>Operator Mappings</h3>
 
       <section id="triple-term-comparison">
         <h3>Triple terms with ORDER BY</h3>
-
-        <p>There is an added item to the ordering of terms when they would not 
-          otherwise be ordered. RDF-star triple terms are placed last (highest).</p>
+        <p>
+          There is an added item to the ordering of terms when they would not 
+          otherwise be ordered. RDF-star triple terms are placed last (highest).
```
TallTed, line 2498:
```suggestion
          There is an item added to the ordering of terms when they would not 
          otherwise be ordered: RDF-star triple terms are placed last (highest).
```

TallTed (changes requested):
See above.

cg-spec/editors_draft.html
afs: I hope you find this better:
```
          When comparing two RDF-star
          triple terms, comparison is by pairwise comparison of the components: first
          compare the subjects, then compare the predicates, then compare the objects.
```

afs: This text is supposed to just giving an outline if the process, with the definition in the box.

I'll reword it - maybe just describing it as "pairwise" is enough.

afs: I'll reword this text. The details are left ot the definition box (e.g. only compare predicates if the subjects are the same).

cg-spec/editors_draft.html
afs: Changed to `result` in all places.

PR #280, opened 2023-06-23
 feat: add negative tests for annotation syntax in nt

jeswr:
@afs Do the manifest.ttl and manifest.jsonld both need to be manually updated in this repo - or is there some kind of automation I should be running to update them?

### DIFF
--- a/tests/nt/syntax/manifest.jsonld
+++ b/tests/nt/syntax/manifest.jsonld
@@ -193,6 +193,30 @@
       "@type": "rdft:TestNTriplesNegativeSyntax",
       "action": "ntriples-star-bad-syntax-4.nt",
       "name": "N-Triples-star - Bad - quoted triple, blank node predicate"
+    },
+    {
+      "@id": "trs:ntriples-star-bnode-bad-annotated-syntax-1",
+      "@type": "rdft:TestNTriplesNegativeSyntax",
+      "action": "ntriples-star-bnode-bad-annotated-syntax-1.nt",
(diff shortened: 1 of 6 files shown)


afs:
manifest.ttl needs to be updated.

A conversion to json-ld would be nice (Turtle can have `#` comments).

gkellogg:
Note that w3c/rdf-tests#95 is intended to automatically update manifest.jsonld and the associated HTML version in each PR from manifest.ttl. This is similar to how it's done in https://github.com/w3c/rdf-canon/blob/main/.github/workflows/generate.yml. In RCH, any time something is touched in the tests or reports directory, it re-builds the dependent files and pushes up a new commit to the PR.

afs:
I've checked the tests and executed manifest.ttl - all OK. So this PR is ready. Thanks!

This is the rdf-star (CG) repo. 

The tests will eventually end up in [rdf-tests](https://github.com/w3c/rdf-tests) but maybe not directly. The RDF -star WG may wish (or even need) to have it's own test area so that it can mark them as the test suite at the end of the working group.

I've emailed the WG to get the process details agreed upon.

https://lists.w3.org/Archives/Public/public-rdf-star-wg/2023Jun/0034.html

gkellogg (approved):

jeswr:
Awesome - I don't have permissions to merge on this repo so @gkellogg @afs you can go ahead and do this yourselves when you're ready.

afs:
The WG still has to work out the over details but for now, merging to the CG repo is OK.

Thanks!

